PR #32, opened 2018-03-08
Adding support for Adal auth, Microsoft Teams support, fixes

patrick-rodgers:
#### Category
- [X] Bug fix?
- [X] New feature?
- [ ] New sample?
- [X] Documentation update?

#### Related Issues

none

#### What's in this Pull Request?

Adds support for adal auth (AdalClient in .@pnp/common), Microsoft teams support (basic), adal client article, small fixes for tooling and banner.

PR #35, opened 2018-03-09
@pnp/graph: Added support for beta endpoint

olemp:
#### Category
- [ ] Bug fix?
- [x] New feature
- [ ] New sample?
- [ ] Documentation update?

#### Related Issues

No related issues.

Added support for /beta endpoint of the graph.

### DIFF
--- a/packages/graph/src/beta.ts
+++ b/packages/graph/src/beta.ts
@@ -1,0 +1,23 @@
+import { GraphQueryable } from "./graphqueryable";
+import { Groups } from "./groups";
+
+/**
+ * Root object wrapping beta functionality for MS Graph
+ *
+ */
+export class Beta extends GraphQueryable {
+
+    /**
+     * Creates a new instance of the Beta class
+     *
+     * @param baseUrl The url or Queryable which forms the parent of this fields collection
+     * @param path Optional additional path
+     */
+    constructor(baseUrl: string | GraphQueryable, path = "beta") {
+        super(baseUrl, path);
+    }
+
+    public get groups(): Groups {
+        return new Groups(this);
+    }
+}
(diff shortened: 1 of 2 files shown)


patrick-rodgers:
Hi @olemp - thanks for this, but have taken a slightly different approach than before. Now I have removed the v1 etc from the fluent api. it is just graph.groups and where needed there is a method to switch the api that is hidden from the user. This comes from the idea that 1) folks don't really care what they are calling and 2) there aren't plans to have a ton of end points. Issue is that change is currently sitting in #32. I will have to go through each of your PRs (thank you for all of them) and see how best to merge once I get #32 merged. Happy to work with you on best way to get these changes in.

olemp:
I see, that's a much better approach indeed!

patrick-rodgers:
hI @advdberg - best to submit feedback like this as an issue so we can properly track it and respond. Comments on closed PRs tend to get missed.

The method I think you are referring to is ".Beta"? That is no longer used as there is no longer a [beta property in the graph export](https://github.com/pnp/pnpjs/blob/dev/packages/graph/src/rest.ts). The way we handle it now is the [setEndpoint method](https://github.com/pnp/pnpjs/blob/dev/packages/graph/src/teams.ts#L68). If you meant why is that method protected it is to hide the implementation detail from the user. The design of the API would hide it so that the caller wouldn't know/care what endpoint they are calling.

So what exactly isn't supported? The correct way to fix it would be to update the library to switch to the beta endpoint when needed. If possible can you submit this as an issue so we can track it? Would be a big help.

advdberg:
@patrick-rodgers thanks for the quick reaction! I will submit a new issue with some more information.

Issue #68, opened 2018-04-12
How to rename SharePoint folder with pnp/sp?


Dzieni:
### Category
- [ ] Enhancement
- [ ] Bug
- [x] Question
- [x] Documentation gap/issue

### Version
@pnp/sp v. 1.0.4

### Question
I need to rename SharePoint folder. How can I do that?

### Observed Behavior
I looked into Folder class source code and found `update()` function. Tried to use it like this:
```javascript
sp.web.getFolderByServerRelativePath('/my/path/here').update({Name: 'newName'})
```
Request is getting passed, but it doesn't affect the folder name.

Thank you in advance for the help.

koltyakov:
Hi Michał,

First of all, thanks for using the lib!

I can see some probable issues related to incorrect usage of `getFolderByServerRelativePath`.
`getFolderByServerRelativePath` expects relative path without `webServerRelativeUrl` part and without `/` in the beginning.
In some cases, it can be handy using `getFolderByServerRelativeUrl`.

Then, folder name can't be updated like this, you got to get folder's item and update `FileLeafRef` property. This is REST API related nuances so they can be omitted in PnPjs's docs.

Here is the example that should make all clear:

```typescript
/**
 * Where `/sites/site` is a `webServerRelativeUrl`
 */
// const folder = pnp.sp.web.getFolderByServerRelativeUrl('/sites/site/DocLib/Folder1');
const folder = pnp.sp.web.getFolderByServerRelativePath('DocLib/Folder1'); // equivalent

folder.getItem()
  .then(item => item.update({ FileLeafRef: 'Folder2' }))
  .then(console.log);
```

Dzieni:
@koltyakov 
Thank you very much!

Two supplementary questions:
1. You mentioned differences between `getFolderByServerRelativePath` and `getFolderByServerRelativeUrl`. I haven't noticed them before because `getFolderByServerRelativePath` works for both absolute and relative paths. Is it an excepted behavior?
[...]

PR #75, opened 2018-04-18
Implementation of folder's moveTo method

koltyakov:
#### Category

- [ ] Bug fix?
- [x] New feature?
- [ ] New sample?
- [ ] Documentation update?

#### Related Issues

mentioned in #68

The PR adds an experimental `moveTo` method for the `Folder` object. The method uses `/_api/SP.MoveCopyUtil.MoveFolder()` API endpoint and allows moving folders within the folders tree.
In comparison to the existing possibility of moving a folder by updating `FileLeafRef` prop, the method allows moving folders on the hight level or a sibling levels (not in the same root folder). Updating FileLeafRef moves folder to the descendant levels and renaming the folder.

@patrick-rodgers, in the implementation I can see some points where I need your opinion on:

- Helper method doesn't mirror API Reference strictly, yet:
  - Being a member method of a Folder might be logical for the library users. We got a folder, then we have the option to move.
  - I'm not very happy with implementing separate `SP.MoveCopyUtil` endpoints for some reasons (redundancy, lack of verbosity, etc.). 
- I didn't find another method as applying `SPHttpClient`, which I know can be probably awkward using inside such methods, also there is some mess with getting urls. Would love hearing about alternatives from you. I would use a new queriable object (or clone and postCore, yet didn't figure out how to deal with custom endpoint here), but can't see a necessity of creating a class for `SP.MoveCopyUtil` at this point.

Please consider the changes from this PR as experimental. I would love to follow the recommendation and applying changes if needed. Also, I'm ok if we will decide to close this PR without merging in case if the implementation looks awkward, etc. :)


wmertens:
We noticed that if there is an error with the from or to paths, the API will still return 200. Is there some header or so that will let the API return an error when source or destination are missing?

koltyakov:
Thanks @wmertens! I have a feeling it's `MoveFolder`'s behavior, I initially tested with `MoveFolderByPath` it was returning the errors I'm sure as I was getting a bunch of them in a progress. Then I switched to `MoveFolder` method just for the reason of making code base smaller (MoveFolderByPath requires larger payload object format). But I'll better switch to the method which responds with an error on errors.

koltyakov:
Humble update: Still didn't have a chance with this PR. Some url patters seemed always return 200 even being incorrect.

patrick-rodgers:
All good, you can merge when ready :)

wmertens:
@koltyakov so you experienced the 200 returns on faulty inputs with `MoveFolderByPath` as well?

StJoshua:
HotFix=[[VAULTED INPUTS]]

StJoshua:
Angels+Humanity=(PURE-TRUTH)

koltyakov:
@wmertens, yup, `200 OK` in response from `MoveFolder` and `MoveFolderByPath` when URL parameters are absolute yet incorrect paths (i.e. non-existing source url or dest path which could not be created).

wmertens:
Maybe this function can optionally verify the paths before doing the move? Because if you do it afterwards, there can be a network issue leading to false negative

koltyakov:
I've merged the PR to let folks already get advantages from using `SP.MoveCopyUtil.MoveFolder`, though was not able to found a way to overcome the situation when an OK responce is send by the API while an actual action failed. For this case, I consider we'll provide some recommendations like: 1. check if source exists, target not exists, 2. use moveTo method, 3. check if target folder exists.

PR #116, opened 2018-05-31
Tagging availability, default value

fthorild:
#### Category
- [ ]  Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?
- [x]  Suggestion?

#### Related Issues

None

I suggest a change for the default parameter isAvailableForTagging to true for the below method:

```javascript
public addTerm(name: string, lcid: number, isAvailableForTagging = true, id = getGUID()): Promise<ITerm & ITermData>
```
My thinking behind it and the reason is to align with the behavior in the UI or CSOM where the same parameter defaults to true.

Really excited about the taxonomy coming in to this initiative, well done!! Thanks!


patrick-rodgers:
Makes sense, thanks for the contribution!

PR #150, opened 2018-06-25
escapedStringToJson more unescape cases

### LINKED ISSUE #148: Get page details when people web part on the page
### Category
- [ ] Enhancement
- [x] Bug
- [ ] Question
- [ ] Documentation gap/issue

### Version

Please specify what version of the library you are using: [ ^1.1.2-0        ]

Able to fetch a page using the latest release. Attempting to re-test a previous bug logged, see below, with latest release 

https://github.com/pnp/pnpjs/issues/125

### Observed Behavior
Am getting an error, see below

Unexpected token ? in JSON at position 372

### Steps to Reproduce
Create a new page and add people web part to the page. Execute code to fetch the page. After executing code below you get the error above

const pageLayoutDetails = await ClientSidePage.fromFile(
        this.props.spRest.web.getFileByServerRelativeUrl(
          this.props.pageLocation));

koltyakov:
#### Category
- [x] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?

#### Related Issues

fixes #148

Fixes some cases with webparts definition parsing from string to JSON.

### DIFF
--- a/packages/sp/src/clientsidepages.ts
+++ b/packages/sp/src/clientsidepages.ts
@@ -254,15 +254,16 @@ export class ClientSidePage extends File {
      * @param escapedString 
      */
     public static escapedStringToJson<T = any>(escapedString: string): T {
+        const unespace = (escaped: string): string => {
+            const mapDict = [
+                [/&quot;/g, "\""], [/&#58;/g, ":"], [/&#123;/g, "{"], [/&#125;/g, "}"],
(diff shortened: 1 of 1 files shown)


patrick-rodgers:
Nice work. I hate that we have to do this and I looked at trying to come up with a cleaner way, but this seems like the best path. Likely we'll have to keep updating this as we find more characters. This is a change from the original release on the server side as well since nothing used to be escaped except the first 4.

PR #151, opened 2018-06-25
Docs minor restyling

koltyakov:

This PR is about the suggestion of minor styles changes for the documentation static site.


koltyakov:
@patrick-rodgers could you please take a look, what do you think of the minor styles changes? This is the concept I shared before.

patrick-rodgers:
Small changes to copy img folder into docs output which makes the header image path relative. Like I said before - really like the changes, thanks for the updates!

Issue #167, opened 2018-07-13
Documentation: pnpjs Class diagram to print?


lrolseth:
- [ ] Enhancement
- [ ] Bug
- [x] Question
- [x] Documentation gap/issue

Greetings, has anybody created a class diagram of pnpjs?  I'm looking for something to print to put on my wall as a visual reference.  Either UML class diagram, mind map, etc.  I know I can just print off the list of classes, I'm particularly looking for a visual representation of them.

patrick-rodgers:
Hi @lrolseth - no, we don't have something like that. If someone from the community wanted to generate a graphic that would be fantastic, but this isn't something we are likely to do ourselves.

PR #183, opened 2018-07-29
Class diagrams in docs

fthorild:
This PR contains a response to issue #167 

I might have gone a bit above and beyond as I generated class diagrams (svg) and added a link in the bottom of each packages index.md (one for graph one for odata etc...) rather than just reply in the issue with the diagram :)

I am more than willing to keep them up to date.

Let me know if I should response in the issue instead


patrick-rodgers:
Great stuff, thanks for this!! Will get this merged shortly, appreciate the work here, this will be a big help for folks I think.

Issue #52, opened 2018-03-19
Web part data version attribute wrong


raymondlittle100:
### Category
- [ ] Enhancement
- [ x] Bug
- [ ] Question
- [ ] Documentation gap/issue

### Version

Please specify what version of the library you are using: [     1.0.3    ]

### Expected / Desired Behavior / Question
Adding a new OOTB SharePoint 'People' web part and configuring via library. Trying to set this to use the 'Descriptive' layout.

### Observed Behavior
Web part always defaults to 'Compact' layout

### Steps to Reproduce
Fetch web part definitions
private getClientSideWebPartDefinitions(): Promise<ClientSidePageComponent[]> 
  { 
    return new Promise<ClientSidePageComponent[]>((resolve) => 
    { 
      resolve(this.props.spRest.web.getClientSideWebParts());
    });  
  }

Create custom interface to replicate OOTB people web part properties, importantly setting layout to 2

private getContactWebPartProperties() :ISharePointPersonContact
  {  
    let contactUsers:IPersonsEntity[] =[];

    const ownerUser = getMetadataItem(this.state.pageMetadata, "KIRA_CP_ContentOwner");
    let ownerContact:IPersonsEntity = {
      id:ownerUser.fieldValue[0].tertiaryText,
      upn:ownerUser.fieldValue[0].secondaryText,
      role:"",
      department: "",
      phone:"",
      sip:""         
    };

    contactUsers.push(ownerContact);
    
    let contactWebPartProps: ISharePointPersonContact=
    {
      layout:2,
      persons:contactUsers      
    };

    return contactWebPartProps;
  }

Find web part definition and add web part setting to properties created in snippet above. 

// find the definition we want, here by id
    // const partDef = parts.filter(c => c.Id.toLocaleLowerCase() === "{"+webPartId.toLowerCase()+"}");
    let partDef = webPartDefinitions.filter(c => c.Id.toLocaleLowerCase() === "{"+webPartId.toLowerCase()+"}");

    /*if its null could be because of issue where some web parts start 
    with '{' and some don't. So first try with '{' then without*/
    if(partDef == null || partDef.length ==0)
    {
      partDef = webPartDefinitions.filter(c => c.Id.toLocaleLowerCase() === webPartId.toLowerCase());
    }
    
    // optionally ensure you found the def
    if (partDef.length < 1) {
        // we didn't find it so we throw an error
        throw new Error("Could not find the web part");
    }

    // create a ClientWebPart instance from the definition
[...]

PR #242, opened 2018-09-14
Added "1.0"  as default value to dataVersion in the "import" function.

simonagren:
Without dataVersion the imported OOTB Web Part doesn't work on the page.

#### Category
- [ X ] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?

#### Related Issues

When importing a Web Part and trying to add it you get errors without supplying a "dataVersion".
So it's better to have 1.0 as a default value.

image

After adding a dataVersion:

image

image

Changed the default value in the import function to be 1.0.

### DIFF
--- a/packages/sp/src/clientsidepages.ts
+++ b/packages/sp/src/clientsidepages.ts
@@ -759,7 +759,7 @@ export class ClientSideWebpart extends ClientSidePart {
         const manifest: ClientSidePageComponentManifest = JSON.parse(component.Manifest);
         this.title = manifest.preconfiguredEntries[0].title.default;
         this.description = manifest.preconfiguredEntries[0].description.default;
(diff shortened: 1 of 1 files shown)


simonagren:
Just some additional observations:
When we add the web part to the page - for instance the NewsWebPart - we can use 1.0, 1.5, 2.0 as dataVersion, that's just fine.
It doesn't matter how many times we edit and save the page via Rest (PnPJS), the dataVersion won't change,

image


But after the webpart is added on the page with whatever dataVersion, a manual edit and publish actually sets the dataVersion to the "correct one".

image

patrick-rodgers:
Thanks for this update, definitely a good catch. Also relates to issue #52 as we can't determine a version from the client currently. Thanks!

PR #243, opened 2018-09-18
renderListData, renderListFormData: OData verbose mode issue fix

koltyakov:
#### Category

- [x] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?

Fixed renderListData, renderListFormData methods result parsing in verbose OData mode.


patrick-rodgers:
So we were just looking for the wrong property name? That's odd, I wonder if it changed or if we had it wrong all this time

koltyakov:
Property names were correct, yet the order to parse a result as a JSON wasn't.
With OData=verbose, the response object has the following shape:

```json
{
  "RenderListData": "[JSONified string]"
}
```

JSON.parse was applied to the object but not a property.
On some stage, JSON.parse was moved before hOP condition.

patrick-rodgers:
It looks like this change broke a test, can you have  look? https://travis-ci.com/pnp/pnpjs/builds/85482936

koltyakov:
Oh, my fault, I missed that the property name was incorrect. Pushed a quick fix.

image

PR #311, opened 2018-10-27
Update issue AppPermissionRequest XML debugging.md

KEMiCZA:
#### Category
- [ ] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [x] Documentation update?

Had an issue where the permissions would not be applied. In code I would always get a HTTP/1.1 403 Forbidden error. First I thought there was an issue in the code, but I could reproduce the same error using the GetAppOnlyAuthenticatedContext method in PnPCore C#.

I found that changing the http://sharepoint/content/sitecollection to http://sharepoint/content/tenant worked fine. They both result in the same permission:

* Let it have full control of this site collection.
* Let it share its permissions with other users.
* Let it access basic information about the users of this site.
* Allow this app to execute search queries on your behalf, ignoring the app's permissions on result items.
* Allow application access to user profiles: Full Control

### DIFF
--- a/packages/documentation/debugging.md
+++ b/packages/documentation/debugging.md
@@ -146,7 +146,7 @@ Now that we have created an add-in registration we need to tell SharePoint what 
 
 ```XML
   <AppPermissionRequests AllowAppOnlyPolicy="true">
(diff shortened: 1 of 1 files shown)


patrick-rodgers:
Thanks for this update!

PR #314, opened 2018-10-29
Build system updates

patrick-rodgers:
#### Category
- [X] Bug fix?
- [X] New feature?
- [ ] New sample?
- [ ] Documentation update?

This PR updates the build system to fix some issues introduced with the move TypeScript project references. Namely intellisense and go to definition are now working as expected. This required a file named tsconfig.json be present in each package directory.

Also this updates the build system code to laregely remove gulp in favor of invoking the cli methods directly for webpack, rollup, and other commands. Additionally things are now controlled centrally from webpack.config.js and rollup.config.js in the root of the package. These changes build on the previous work and enable:

- running tsc -b from the root
- running webpack from the root
- running rollup -c from the root folder

Gulp commands remain unchanged and function as before. Builds are now somewhat faster but also more predictable due to centralized configuration and calculation of various properties from the tsconfig.json files that previously had to be supplied. Idea behind this effort is to remove as much custom logic as possible and focus on configuration of the tools.


patrick-rodgers:
@pedro-pedrosa , @koltyakov , @KEMiCZA  - you all are doing a lot of work in the library currently. If you have a chance could you pull this PR and do a quick sanity check. I have done a lot to update the build system under the hood such that the changes shouldn't affect anything anyone is doing - but want to avoid the "it works on my machine" moment if possible. Thanks!

KEMiCZA:
@patrick-rodgers 
I've tested the following gulp tasks and they all work fine: build, clean & clean-build

Then inside the packages folder I was able to successfully execute the folowing cli commands:
- tsc -b

For the following commands I have some remarks:
- rollup -c & webpack: for these I had to gulp build first. If you only did the build with tsc -b inside the packages folder, not all the packages would be built (for example packages-es5) and an error would occur. I assume this is normal?

Also I had to install the CLI's manually first using npm install ... -g. But I assume this is also normal?

patrick-rodgers:
@KEMiCZA - Thanks for having a look!

> * rollup -c & webpack: for these I had to gulp build first. If you only did the build with tsc -b inside the packages folder, not all the packages would be built (for example packages-es5) and an error would occur. I assume this is normal?

Yes, that is by design. You would have to run tsc twice from the command line. gulp build does this for you, as well as a a few other things. Same with gulp package, running webpack and rollup isn't enough to produce the final package structure in dist. The goal was however for those config files to control everything as opposed to some of the settings getting magicly set deep in the buildsystem code.

> 
> Also I had to install the CLI's manually first using npm install ... -g. But I assume this is also normal?

Yes, unless you use the gulp commands which ref the files in the local node_modules. If you don't try to use those directly then you shouldn't need to do any additional installs beyond the npm install when you download the library.

KEMiCZA:
Great thanks for the info Patrick.

patrick-rodgers:
Agree, the last few weeks have been really exciting to see the jump in contributions. It really shows in the changelog additions!

PR #328, opened 2018-10-31
Fixed typo in SearchPropertyValue interface (Intval -> IntVal)

luismanez:
#### Category
- [X] Bug fix?

Just fixed a minor type in SearchPropertyValue interface. This caused an issue when you want to use that property in a Search query, as this generated JSON is not OK to SharePoint search service, as is Case sensitive, so it fails with Intval, but works with IntVal (capital V)

```json
 {
   "Name": "ListItemId",
   "Value": {
     "QueryPropertyValueTypeIndex": 2,
     "Intval": 16
    }
}
```

Thanks,

Luis.

### DIFF
--- a/packages/sp/src/search.ts
+++ b/packages/sp/src/search.ts
@@ -662,7 +662,7 @@ export interface SearchProperty {
 export interface SearchPropertyValue {
     StrVal?: string;
     BoolVal?: boolean;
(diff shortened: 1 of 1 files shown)


patrick-rodgers:
Nice catch, thanks!

PR #352, opened 2018-11-08
Added IRefiner interface over ResultTable.Refiners.

pljfdi:
Should make them easier to return to results

#### Category
- [ ] Bug fix?
- [x] New feature?
- [ ] New sample?
- [ ] Documentation update?

#### Related Issues

You cannot easily return the Refiners from the SearchResults, without having to an entire ResultsTable. Which is a little wasteful.
This is a simple interface, that allows you to return only the refiners array explicitly.

sp/Search.ts added interface `IRefiner` and altered the `ResultTable` interface changing ResultTable.Refiners to be `IRefiner`

### DIFF
--- a/packages/sp/src/search.ts
+++ b/packages/sp/src/search.ts
@@ -619,12 +619,16 @@ export interface ResultTableCollection {
     SpecialTermResults?: ResultTable;
 }
 
+export interface IRefiner {
+    Name: string;
+    Entries: { RefinementCount: string; RefinementName: string; RefinementToken: string; RefinementValue: string; };
+}
 export interface ResultTable {
     GroupTemplateId?: string;
     ItemTemplateId?: string;
     Properties?: { Key: string, Value: any, ValueType: string }[];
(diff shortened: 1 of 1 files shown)


patrick-rodgers:
Thanks, this should be a help to folks!

bcage29:
@pljfdi @patrick-rodgers this change causes a bug. The new interface is missing brackets at the end to indicate it is an array.

{ RefinementCount: string; RefinementName: string; RefinementToken: string; RefinementValue: string; }[];

pljfdi:
The new interface isn't an array type object, and isn't meant to be. It's just the layout of what *a* refiner is. All the `IRefiner` does is alias a bunch of fields behind a name, so that you can more easily get the Refiner results from a Search Result Table.
The implementation in `ResultTable` is an array, as it's that object that returns multiple Refiners.

Have you got a code example where it would cause a break, or a bug?

patrick-rodgers:
If there is a bug please open an issue so we can properly track it and ensure it is resolved. Thanks!

cslecours:
I have opened a bug and made a pull request.

PR #399, opened 2018-12-02
add window.performance to prevent collisions

AustinBreslinDev:
#### Category
- [X] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?

#### Related Issues
Not raised yet, checked references to method and found none. Could possibly be removed from code base.

The getGuid method relies on the uniqueness of the generated number from the Date object, but during simple tests it is apparent that this is not always the case.

Tested in Firefox and Chrome, the uniqueness in Firefox stays the same.

Run this in the Chrome Browser Developer Console
```
for(var i = 0; i < 50; i++) {
    console.log(Date.now());
}
```
You could/should see something like this

image

Run this
```
for(var i = 0; i < 50; i++) {
    console.log(performance.now());
}
```

You will see much more lines in the output, thus uniqueness
image

### DIFF
--- a/packages/common/src/util.ts
+++ b/packages/common/src/util.ts
@@ -75,10 +75,14 @@ export function getRandomString(chars: number): string {
  * Gets a random GUID value
  *
  * http://stackoverflow.com/questions/105034/create-guid-uuid-in-javascript
+ * https://stackoverflow.com/a/8809472 updated to prevent collisions.
  */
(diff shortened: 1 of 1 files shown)


patrick-rodgers:
Thanks. Not sure how much of an issue this would be, but nice to get the latest updates. Appreciate you submitting the PR!

PR #434, opened 2018-12-20
added more parameters to list subscription update method

tavikukko:
#### Category
- [ ] Bug fix?
- [x] New feature?
- [ ] New sample?
- [ ] Documentation update?

While refactoring the list subscription **add** method, realised that when **updating** the list subscription, you can update all the parameters: **expirationDate** and **notificationUrl** and **clientState**, before it was only possible to update the **expirationDate**. So I made all the parameters optional in the update method. I was wondering should we create methods for each case?

Also this should be backward compatible with earlier PnPjs versions.

### DIFF
--- a/packages/sp/src/subscriptions.ts
+++ b/packages/sp/src/subscriptions.ts
@@ -61,7 +61,7 @@ export class Subscription extends SharePointQueryableInstance {
       };
 
       if (expirationDate) {
-        postBody.expirationDate = expirationDate;
+        postBody.expirationDateTime = expirationDate;
       }
 
       if (notificationUrl) {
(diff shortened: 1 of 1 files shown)


tavikukko:
Did also some tests. With this update, if you call it without any parameters (which is possible as all parameters are now optional), the actual rest call is not failing, it jus and it doesn't update anything with empty body, nice.

patrick-rodgers:
Not sure how it will merge but it looks like this branch doesn't have the update to the add method you submitted in another PR. Do you want to refresh this branch to ensure it has the latest before I merge? Would hate to overwrite your other change.

tavikukko:
I think I got it merger all right.

Issue #69, opened 2018-04-13
parentWeb endpoint missing


jimmywim:
### Category
- [*] Enhancement
- [ ] Bug
- [ ] Question
- [ ] Documentation gap/issue

### Version

Please specify what version of the library you are using: [1.0.4]

### Expected / Desired Behavior / Question
ParentWeb is available as an endpoint on _api/web but not yet implemented in code

### Observed Behavior
endpoint is missing from Web object

Is this endpoint slated to be added anytime soon? 

If not, is it possible to use whatever http fetcher pnp uses under the hood to cheekily call web/parentWeb and cast the returned data into a pnp Web object?

koltyakov:
Hi @jimmywim,

Not an issue to add this enhancement. Actually, I already did it in a local branch and in a process of testing before PR.

Thanks for enhancement suggestion offer!

koltyakov:
After some tests:

`ParentWeb` endpoint is not the same as having an instance of the web.

And my initial idea of adding:

```typescript
    /**
     * Gets this web's subwebs
     *
     */
    public get webs(): Webs {
        return new Webs(this);
    }
```

to Web class failed on deep testing.

Let me elaborate a bit on this.

My idea was to adding `sp.web.parentWeb` returning an instance of `Web`. But `parentWeb` is just a set of properties and can't be chained with other usual endpoints. I.e. `sp.web.parentWeb.lists` will fail, `sp.web.parentWeb.select('Url')` won't return the prop.

In existing code base we can actually deal with `parentWeb` in the following way:

```typescript
pnp.sp.web.select('ParentWeb').expand('ParentWeb').get().then(console.log);
```

So... I'm not sure, do we need parentWeb as an additional get method? Any thoughts?

jimmywim:
That method appears to get some data about the parent Web but it's not of type `Web` in the code base, so doesn't have all the methods & properties on it (which is useful, and what I currently need as I'm in a recursive method trying to pull out Publishing's available page layouts - agreed that's not something to deal within Modern sites but the ability to get the full Web object for the parent might still be useful).

The object that comes back does have Id though, so we can use `Site.openWebById()` on it.

koltyakov:
A web object can be constructed using `const web = new Web([web absolute url])`.
In most situations, I would use existing web URL and pop last path part out and build parent web object with full API support.

Or, as you said, site.openWebById is the case too:

```typescript
(async () => {

[...]

PR #442, opened 2018-12-27
fix getParentWeb method

### LINKED ISSUE #441: GetParentWeb method throwing error
### Category
- [ ] Enhancement
- [x] Bug
- [ ] Question
- [ ] Documentation gap/issue

@pnp/sp: [    "1.2.7"     ]

SharePoint version: [    SPFx     ]

I'm trying to use the getParentSite() method.

### Observed Behavior
imagem

It's throwing this error. I have set up the context as such:

  sp.setup({
    spfxContext: context  
  });

Context is this.context coming from the onInit method of my ApplicationCustomizer. (SPFx Extension)

### Steps to Reproduce
Just try using the method on SPFx.

Thanks!

tavikukko:
#### Category
- [x] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?

#### Related Issues

fixes #441

There was error constructing new Site. Also if Web is rootWeb, it does not have ParentWeb, so made it to return null in that case.

### DIFF
--- a/packages/sp/src/webs.ts
+++ b/packages/sp/src/webs.ts
@@ -111,7 +111,7 @@ export class Web extends SharePointQueryableShareableWeb {
      */
     public getParentWeb(): Promise<{ data: any; web: Web }> {
         return this.select("ParentWeb/Id").expand("ParentWeb").get()
-            .then(({ ParentWeb }) => new Site(this.toUrlAndQuery().split("/_api")[0]).openWebById(ParentWeb.Id));
+            .then(({ ParentWeb }) => ParentWeb ? new Site("").openWebById(ParentWeb.Id) : null);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


packages/sp/src/webs.ts
```diff
@@ -111,7 +111,7 @@ export class Web extends SharePointQueryableShareableWeb {
      */
     public getParentWeb(): Promise<{ data: any; web: Web }> {
         return this.select("ParentWeb/Id").expand("ParentWeb").get()
-            .then(({ ParentWeb }) => new Site(this.toUrlAndQuery().split("/_api")[0]).openWebById(ParentWeb.Id));
+            .then(({ ParentWeb }) => ParentWeb ? new Site(this.parentUrl).openWebById(ParentWeb.Id) : null);
```
patrick-rodgers: Why not here just get the ParentWeb/Url and return a new web that way? I am sure I wrote the old code, but looking at it I think we can come up with a better way.

tavikukko: Here a previous discussion about the implementation #69

patrick-rodgers: Oh right, ParentWeb isn't a web instance but WebInformation instance. Got it.

Issue #457, opened 2019-01-15
Content type id ignored when using sp.web.contenttypes.add


olemp:
### Category
- [ ] Enhancement
- [x] Bug
- [ ] Question
- [ ] Documentation gap/issue

### Version

Please specify what version of the library you are using: [1.27]

Please specify what version(s) of SharePoint you are targeting: [Sharepoint Online]

### Expected / Desired Behavior / Question
When using `sp.web.contenttypes.add`, I excepted the created content type to have the same StringId that was specified in the request.

### Observed Behavior
When using `sp.web.contenttypes.add`, the created content type gets a random StringId/ContentTypeID.

### Steps to Reproduce
Try the following function call:

`web.contentTypes.add("0x01003376E0D552D746EBBB008B9E999064E1
", "pnpjs content type", "", "Custom Content Type")`

Check if the content type id of the created content type `pnpjs content type` is `0x01003376E0D552D746EBBB008B9E999064E1`.

olemp:
Is there a "limitation" in the SharePoint APIs @patrick-rodgers?

koltyakov:
@olemp, as far as I could recall, content type ID is ignored with the REST API add method by itself. In other words, something which can't be fixed within the library but an external bug of the API.

olemp:
@koltyakov Yes, seems like that's the case.

patrick-rodgers:
Did a little more digging here and the object you are supposed to pass to the create method (post to the contenttype collection) is SPContentTypeEntityData and does have a ParentContentTypeId property - BUT you can't actually use this class as it doesn't resolve. So this is actually a double bug in content type creation IMO as you can't specify the Id directly, and you can't use this entity data type to specify the parent id either. Long story short content types are very broken in REST API. I hate to say it, but for now working with content types is better supported in CSOM.

```C#
[ClientNS.ClientCallableType(
    Name="ContentTypeEntityData",
    ServerTypeId="30e94405-dfd3-472e-b5d7-637ce8e9586c",
    ValueObject=true,
    ClientLibraryTargets=ClientNS.ClientLibraryTargets.IntrinsicRESTful
    )]
internal class SPContentTypeEntityData
{
    [ClientNS.ClientCallableAttribute]
    public string ParentContentTypeId { get; set; }

    [ClientNS.ClientCallableAttribute]
    [ClientNS.ClientCallableConstraint(Type = ClientNS.ClientCallableConstraintType.NotNull)]
    public string Name { get; set; }

[...]

PR #459, opened 2019-01-15
Remove id as param from contenttypes.add since it's not supported by the SP APIs #457

olemp:
#### Category
- [?] Bug fix
- [ ] New feature
- [ ] New sample
- [ ] Documentation update

#### Related Issues

https://github.com/pnp/pnpjs/issues/457

### DIFF
--- a/packages/sp/src/contenttypes.ts
+++ b/packages/sp/src/contenttypes.ts
@@ -40,7 +40,6 @@ export class ContentTypes extends SharePointQueryableCollection {
     /**
      * Adds a new content type to the collection
      *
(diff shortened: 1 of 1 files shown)


koltyakov:
Thanks! Hovewer... 

No matter that there is an issue with ContentType ID in REST API. Just removing it from add method in PnPjs will introduce a compatibility issue for those who already use the method. Other words it's a braking change that can affect others. IMO, we shouldn't just change the parameters interface but probably document API behavior nuances keeping parameters fingerpring as it is now.

What do you think guys? @olemp, @patrick-rodgers

patrick-rodgers:
Going to close this as we can't merge it as-is.

PR #488, opened 2019-02-04
Added remove() to members and owners

simonagren:
#### Category
- [ ] Bug fix?
- [X] New feature?
- [ ] New sample?
- [ ] Documentation update?

Small PR with an addition of removing members and owners, essentially from a group.
Was thinking about how we should document this @patrick-rodgers?

Since there's no ducmentation of groups, should this be a part of groups or as something separate? I can fix that later

/Simon

### DIFF
--- a/packages/graph/src/members.ts
+++ b/packages/graph/src/members.ts
@@ -31,7 +31,14 @@ export class Members extends GraphQueryableCollection<IMember[]> {
     }
 }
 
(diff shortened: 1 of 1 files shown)


patrick-rodgers:
We can always add a page for groups. Thanks as always for your contributions!

simonagren:
@patrick-rodgers I could make some documentation on groups later.

Have a few PR's coming up for the next version as well :)

PR #741, opened 2019-06-21
possible fix #740

### LINKED ISSUE #740: CreateGroupEx started giving siteUrl empty in response
Started seeing the issue again (2nd time since January). CreateGroupEx endpoint accepts parameters, creates group but it won't create SPO site yet. As a result it returns null for siteurl in the response. After a very long time (sometimes 15-45 mins) the site in SharePoint gets created.

####Which PnP repository should you use to report the issue?####
Any version of the PnP does the same which uses this CreateGroupEx to create sites.

**Category**
[ x] Bug
[ ] Enhancement

**Environment**
[ x] Office 365 / SharePoint Online
[ ] SharePoint 2016
[ ] SharePoint 2013

**Expected or Desired Behavior**
Today if we like to create Office 365 Groups via SharePoint online then we are using CreateGroupEx api. This api is supposed to provision group resources and also a SharePoint site collection. As a result it should return group id and site url in the response.

**Observed Behavior**
At the moment when we call this API we see that group is being created and returns only groupid in the response. We see SharePoint site is not created yet, it takes long time to provision. Which resulted the siteUrl as null in the response for API.

**Steps to Reproduce**
Call the rest api CreateGroupEx with parameters, the request will be 200. But actually only group is being created, but site is not. siteurl is showing null in the response.

Please let me know if anyone else is using this API and any other recommendations to create groups via end point.

KEMiCZA:
#### Category
- [x] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?

#### Related Issues

fixes #740

Possible fix, let me know what you think?

### DIFF
--- a/packages/sp/src/site.ts
+++ b/packages/sp/src/site.ts
@@ -215,7 +215,7 @@ export class Site extends SharePointQueryableInstance {
      * @param owners The Owners of the site to be created     
      */
 
-    public createModernTeamSite(
+    public async createModernTeamSite(
         displayName: string,
         alias: string,
         isPublic = true,
@@ -224,7 +224,7 @@ export class Site extends SharePointQueryableInstance {
         classification = "",
         owners?: string[],
         hubSiteId = "00000000-0000-0000-0000-000000000000",
-    ): Promise<void> {
+    ): Promise<IGroupSiteInfo> {
 
         const postBody = jsS({
             alias: alias,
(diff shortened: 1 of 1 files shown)


patrick-rodgers:
Just getting a chance to review this. If I understand the linked issue the service at times delays the creation of the underlying group (which is a known thing). And also if I am not wrong this method does not work with app only tokens (as noted in the comments). Meaning if it takes sometimes 20 minutes to create one of these sites, we are expecting someone to site with a browser window open for 20 minutes? If that is the case I am not sure we need to add this complexity since this doesn't seem like a realistic scenario. What am I missing?

This is not a criticism of the code in the PR, that all looks fine. I am more trying to understand when it would practically be used within this library?

KEMiCZA:
@patrick-rodgers 
I agree, if it can take up to 20 minutes, then it does not sound that good to keep polling the site status ourselves.
What about providing the GetSiteStatus endpoint as a method in the library? This method would then allow the developer to poll for the site status.

patrick-rodgers:
Yes, adding the method might be the better way, then folks can implement their own polling if needed. We could provide sample code in the docs for that. What I am worried about is folks use this method, don't know that it might take 20 minutes, and end up writing apps that just seem to "hang" for no reason to the users.

KEMiCZA:
Closing this one for now. I'll see when I get some free time to add the GetSiteStatus method.

PR #773, opened 2019-07-10
SP queryable: fix alias matching regex

### LINKED ISSUE #772: Cannot alias name with embedded apostrophe
### Category
- [x] Bug

### Version

Please specify what version of the library you are using: [    1.3.2     ]

Please specify what version(s) of SharePoint you are targeting: [   no idea      ]

*If you are not using the latest release, please update and see if the issue is resolved before submitting an issue.*

The changelog for 1.3.3 does not include anything that might fix this

aliases should always work

### Observed Behavior

When using the following function

```js
export const getFolderByUrl = (url) => {
    const web = new Web(siteUrl)
	// Note: this uses an alias to shorten the path. Make sure they don't clash with other aliases
	return web.getFolderByServerRelativeUrl(
		`!@x1::${
			// TEMP workaround for pnpjs issue #313
			url.replace(/'/g, "''")
		}`
	)
}
```

with the url `"/pathToSite/Shared Documents/blabla/diploma's restaurateurs"`, it cuts the url at the apostrophe and creates the nonsense SP URL `https://sharepointhost/pathToSite/_api/web/getFolderByServerRelativeUrl(@x1%27s%20restaurateurs%27)/listItemAllFields?@x1=%27/pathToSite/Shared%20Documents/blabla/diploma%27`

it seems to me that if an argument starts with `/^!@[a-zA-Z0-9]+::/`, the rest of the argument is the string to alias?

wmertens:
The alias matching regex fails for values with embedded apostrophes, even if they are escaped by doubling them.

#### Category
- [x] Bug fix?

#### Related Issues

fixes #772 for Chrome and Node.js

alias matching happens after URLifying, so embedded apostrophes need to have been escaped by doubling them. This changes the regex so embedded apostrophes are part of the name

HOWEVER. This uses lookbehind, which only works in chrome and nodejs. So it's not really the perfect solution.

What would help is that all arguments that will become aliases will have the `'` changed to `%27%27` but then that can't be double-encoded as will happen in v2.

### DIFF
--- a/packages/sp/src/sharepointqueryable.ts
+++ b/packages/sp/src/sharepointqueryable.ts
@@ -88,7 +88,8 @@ export class SharePointQueryable<GetType = any> extends ODataQueryable<SPBatch, 
 
         const aliasedParams = new Map<string, string>(this.query);
 
-        let url = this.toUrl().replace(/'!(@.*?)::(.*?)'/ig, (match, labelName, value) => {
+        // This regex uses negative lookbehind and lookahead to include escaped apostrophes in the value
+        let url = this.toUrl().replace(/'!(@.*?)::(.*?)(?<!')'(?!')/ig, (match, labelName, value) => {
             Logger.write(`Rewriting aliased parameter from match ${match} to label: ${labelName} value: ${value}`, LogLevel.Verbose);
             aliasedParams.set(labelName, `'${value}'`);
             return labelName;


patrick-rodgers:
I can't merge a change that only works in one browser. In the future please also target your PRs to dev branch instead of master. Going to close this without merging due to solution limitations.

wmertens:
@patrick-rodgers 	right, but it is still a problem, right? Or is the workaround as mentioned at #772 the official approach?

patrick-rodgers:
Please see my comments in the issue.

PR #808, opened 2019-08-05
Removed .only() from social sub-module test file

RamPrasadMeenavalli:
#### Category
- [x] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?
One of the test cases in social are having .only which is getting executed when we develop/test the test scripts for other sub modules. Removed .only() form the test case of social sub module.


patrick-rodgers:
Thanks for catching that!

PR #833, opened 2019-08-28
fix Folder.moveTo

nsoeth:
#### Category
- [ x ] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?

#### Related Issues

fixes Folder.moveTo()

The function Folder.moveTo() used the .toUrl() function, which returned a relative url, so it was not possible to get the host url. Because of this, it returned a 404 error. Replaced it with the value of "odata.id" which includes the absolute url.


patrick-rodgers:
Thanks for catching this and submitting a fix!

PR #846, opened 2019-09-11
getCurrentUserEffectivePermissions low-privileged users fix

### LINKED ISSUE #259: getCurrentUserEffectivePermissions Access Denied for user with read access
### Category
- [ ] Enhancement
- [x] Bug
- [x] Question
- [ ] Documentation gap/issue

### Version

Please specify what version of the library you are using: [  1.2.1 ]

I am trying to use the above function against a list to get users permissions, then check if they have AddItems rights with the hasPermissions function. I am testing with a user that has read access to a site, including the list I am checking against, and I get:

"Error making HttpClient request in queryable: [403]  ::> {"responseBody":{"error":{"code":"-2147024891, System.UnauthorizedAccessException","message":{"lang":"en-US","value":"Access denied. You do not have permission to perform this action or access this resource."}}},"responseHeaders":{}}"

My function call looks like this:

 const web = new Web(webUrl);
    return new Promise<boolean>((resolve, reject) => {
      web.lists.getByTitle(listTitle).getCurrentUserEffectivePermissions()
        .then(res => {
          if(web.hasPermissions(res, PermissionKind.AddListItems)){
            resolve(true);
          }
          else{
            resolve(false);
          }
        })
        .catch(e =>{
          console.log(e);
          reject(e)
        })
    })

I've tried the same against the web instead of the list and get the same result. Increasing the test user to edit results in same error, but giving him full control makes it work. Should I be able to use this function with a user that has only read access? I am converting an older app that was making a call to /EffectiveBasePermissions and this was working, I have not tried it in the newer app because I'd like to use this library to make the call. Thanks!

koltyakov:
#### Category
- [x] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?

#### Related Issues

fixes #259 

Changes the way of requesting `EffectiveBasePermissions` for low-privilege users by consuming `/EffectiveBasePermissions` endpoint except for getting user effective permissions by a login.

All credits go to @paulryan. Thanks!

### DIFF
--- a/packages/sp/src/sharepointqueryablesecurable.ts
+++ b/packages/sp/src/sharepointqueryablesecurable.ts
@@ -39,12 +39,10 @@ export class SharePointQueryableSecurable extends SharePointQueryableInstance {
      * Gets the effective permissions for the current user
      */
     public getCurrentUserEffectivePermissions(): Promise<BasePermissions> {
-
-        // remove need to reference Web here, which created a circular build issue
-        const w = new SharePointQueryableInstance("_api/web", "currentuser");
-        return w.configureFrom(this).select("LoginName").get<{ LoginName: string }>().then(user => {
-
-            return this.getUserEffectivePermissions(user.LoginName);
+        const q = this.clone(SharePointQueryable, "EffectiveBasePermissions");
+        return q.get<any>().then(r => {
+            // handle verbose mode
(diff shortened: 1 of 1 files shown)


patrick-rodgers:
Thanks!

PR #913, opened 2019-11-06
Added functionality to set mailnickname when creating a team

frnk01:
Group names may often consist of special characters, spaces and other characters which is not supported in an email address. This presents the need to be able to set a mailNickname different from the group name.

#### Category
- [x] Bug fix?
- [x] New feature?
- [ ] New sample?
- [X] Documentation update?

#### Related Issues

If creating a team with a complex name which contains special characters, the current implementation throws an error due to limitations in the mailNickname.

Added mailNickname to the teams.create and teams.clone functions, and updated documentation.

### DIFF
--- a/packages/graph/docs/teams.md
+++ b/packages/graph/docs/teams.md
@@ -37,7 +37,7 @@ Then create
 ```TypeSCript
 import { graph } from "@pnp/graph";
 
-const createdGroupTeam = await graph.teams.create('Groupname', 'description', 'OwnerId',{ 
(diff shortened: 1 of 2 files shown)


patrick-rodgers:
Thanks! Typically this would be a breaking change and we shouldn't change the method signature - but this isn't a widely used method (I don't think) so will slide it in and call it out in the changelog. Thanks for taking the time to submit this PR and helping us improve the library!

frnk01:
Where do you want new feature to be submitted? dev or dev-v2 branch?

Issue #914, opened 2019-11-07
Underlying SPFx token provider occasionally hangs without resolving or rejecting


TazzyMan:
### Category
- [ ] Enhancement
- [x] Bug
- [ ] Question
- [ ] Documentation gap/issue

### Version
@pnp/pnpjs ~1.3.3
SharePoint Online - SPFx

I didn't update yet, but based on the source code I can see the issue isn't resolved yet

### Expected / Desired Behavior / Question
Get token from SharePoint and either fail or succeed

### Observed Behavior
Token is requested and promise never returns, killing the promise chain alltogether

### Steps to Reproduce
When entering the page from a link in Modern sharepoint (not a full page refresh) or when all cookies etc are cleared. Probably a race condition in the Modern SharePoint page, but I would at least expect an error in that case. The source code indicates it doesn't handle errors well.
The specific code is below:

```typescript 
    /**
     * Gets an AAD token for the provided resource using the SPFx AADTokenProvider
     * 
     * @param resource Resource for which a token is to be requested (ex: https://graph.microsoft.com)
     */
    public getToken(resource: string): Promise<string> {

        return this.context.aadTokenProviderFactory.getTokenProvider().then(provider => {

            return provider.getToken(resource);
        });
    }
```

patrick-rodgers:
Hi @TazzyMan - if I am understanding you correctly the issue is contained within the SPFx aadTokenProviderFactory? If so there isn't much we can do to fix that, but you could report that issue to the [sp-dev-docs issues list](https://github.com/SharePoint/sp-dev-docs/issues).

That being said, we could add a timeout here to produce an error if we don't get back a token within some amount of time - would that help?

TazzyMan:
Hello Patrick,

As a matter of fact I already resolve the issue with a timeout, but there are cases where it takes up to 70 retries to get that token, which each a 1000ms delay, so I'd rather have the exception.
When I look at the code I miss a .catch() in both the call to getTokenProvider() and getToken() and there are no means of rejecting the promise. I didn't have time to test it yet, so that I could do a pull request but I would expect the code to be like below:

`
    public getToken(resource: string): Promise<string> {
        return new Promise<string>((resolve, reject) => {
            this.context.aadTokenProviderFactory.getTokenProvider()
             .then((provider:any) => {
                provider.getToken(resource)
                .then((token:string) => {
                    resolve(token);
                })
                .catch((error:string) => {
                   reject(error);
                });
            ).catch((error:string) => {
[...]

PR #915, opened 2019-11-08
fix dead promise & handle token provider exception

KEMiCZA:
#### Category
- [x] Bug fix?

#### Related Issues
#914

Fixes a dead promise in the adal getToken method as mentioned in #914

### DIFF
--- a/packages/common/src/adalclient.ts
+++ b/packages/common/src/adalclient.ts
@@ -218,11 +218,8 @@ export class SPFxAdalClient extends BearerTokenFetchClient {
      * 
      * @param resource Resource for which a token is to be requested (ex: https://graph.microsoft.com)
      */
-    public getToken(resource: string): Promise<string> {
-
-        return this.context.aadTokenProviderFactory.getTokenProvider().then(provider => {
-
-            return provider.getToken(resource);
-        });
+    public async getToken(resource: string): Promise<string> {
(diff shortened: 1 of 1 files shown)


koltyakov:
Just wondering what is the difference?
The code after the change in regards to promises flow is just the same and equivalent of the previous implementation, but without async/await. The resulted code in both cases will behave the same, yet we considered to prefer async/await syntax.

KEMiCZA:
My bad, I thought since we are returning .then, and we don't catch the exception, that the exception in the getTokenProvider method would not be handled?
```typescript
 return this.context.aadTokenProviderFactory.getTokenProvider().then()...
```
Somehow for me it seems more clean if using the await for this call, since it would throw an exception. But if there's no difference, then I'm mistaken and also confused why the exception is not thrown as @TazzyMan explained.

koltyakov:
With async/await if `getTokenProvider` fails the whole method rejects the Promise, so there is no need in wrapping with try/catch for example. Maybe `getToken` "hangs" somewhere not being caught for example?

KEMiCZA:
Yeah you're right about the async/await and then/this doing the same thing, same as Patrick said. By bad. Was a little confused. Going to close this PR.

koltyakov:
No problem and thank you for assistance, as always a lot appreciated!

PR #982, opened 2019-12-30
v2 - fix openWebById

tavikukko:
#### Category
- [x] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?

When opening web using **openWebById**, all follow-up rest queries had double _apiWeb/_api/web in url.
```
import { sp } from "@pnp/sp/presets/all";

sp.site.openWebById('00000000-0000-0000-0000-000000000000').then(w => {
   console.log(w.web.toUrlAndQuery())
   //output: https://contoso.sharepoint.com/sites/myweb/_api/Web/_api/web
})
```
This PR fixes that issue

### DIFF
--- a/packages/sp/sites/types.ts
+++ b/packages/sp/sites/types.ts
@@ -113,7 +113,7 @@ export class _Site extends _SharePointQueryableInstance {
         const data = await spPost(this.clone(Site, `openWebById('${webId}')`));
         return {
             data,
-            web: Web(odataUrlFrom(data)),
+            web: Web(odataUrlFrom(data).replace(/_api\/web\/?/i, "")),
         };
     }
 


KEMiCZA:
Using the **extractWebUrl** method also works. Not sure what is prefferable here. Version 1 also had a *Web.fromUrl* static method https://github.com/pnp/pnpjs/blob/639228f3a84f4f38912c1b5c8db0032286cffea8/packages/sp/src/webs.ts#L91-L100

Not sure if we need the static *Web.fromUrl* method in v2 since it was only referenced in the openWebById Site method.

patrick-rodgers:
I did it a little differently using an existing utility method - but also needed to patch the extract odata url method as there was a bug there 😐. Not sure how long it had been there. I also added a test to ensure that the web coming from open by id is chainable. Thanks!

PR #990, opened 2020-01-02
[V2] fixed SPFx onpremise docs & setup for sp

### LINKED ISSUE #978: Can't use PnPJS 2.0 targeting SPFx 1.4.1 (SPO and SharePoint 2019)
### Category
- [ ] Enhancement
- [x] Bug
- [x] Question
- [ ] Documentation gap/issue

### Version

Please specify what version of the library you are using: [2.0.0]

Please specify what version(s) of SharePoint you are targeting: [2019, SPO]

I should be able to use SPFx 1.4.1 "at least" as the foundation for my solution as it could run on both SP2019 and SharePoint Online

### Observed Behavior
When using SPFx 1.4.1 as our foundation for a solution to ensure it works on both Online and SP2019, it's impossible to initialize PnPJS as it requires the `aadTokenProviderFactory` object as part of the context, which is not the case for SPFx 1.4.1 for Web Parts and Extensions.

### Steps to Reproduce
Just create a new solution targeting SP2019 and add PnPJS to the solution, you'll quickly see the red squigglies appear in VS Code.

image

KEMiCZA:
#### Category
- [x] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [x] Documentation update?

#### Related Issues

fixes #978

Adjusted documentation for on-premises SPFx typescript issue & added **any** as type for the setup function.


packages/sp/rest.ts
```diff
@@ -33,7 +33,7 @@ export class SPRest {
      * 
      * @param config The SharePoint configuration to apply
      */
-    public setup(config: SPConfiguration | ISPFXContext) {
+    public setup(config: SPConfiguration | ISPFXContext | any) {
```
patrick-rodgers: I am not sure about typing this param as any? Why not mark the property as optional in ISPFXContext? Putting any here makes it seem like you can just pass us anything and we will somehow figure it out when there are really only two cases covered. Thoughts?

KEMiCZA: I've made some changes, and I agree. Let me know if the latest changes are good. I tested these changes on a SP2019 SPFx project. I also think we can remove both aadInfo & legacyPageContext from the pageContext object. I don't see any references to those in the pnpjs library?

patrick-rodgers: Yes, we can probably drop them. We used to use them but I think we stopped after we introduced the latest graph token stuff a while ago.

patrick-rodgers:
Thanks for the updates - I went ahead and cleaned up the interface based on our discussion. Thanks!!

PR #1025, opened 2020-01-22
fix for #1024

### LINKED ISSUE #1024: Graph : how to list all members of a Group ?
# Category
- [x] Question

### Version
Please specify what version of the library you are using: [    Latest     ]
Please specify what version(s) of SharePoint you are targeting: [    SPO    ]

### Question

I want to perform the equivalent of this REST api call :
GET https://graph.microsoft.com/v1.0/groups/{id}/members

I haven't found the way to do it with pnpjs graph api.

`graph.groups.getById(<id>).members` gives me an IMembers that I can't just "get()".
It is a _GraphQueryableCollection, but even having a look at the pnpjs code, I don't find just how to retrieve it and parse it .

Thanks for your help !

KEMiCZA:
#### Category
- [x] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?

#### Related Issues

fixes #1024

Seems like nothing was getable in graph api.

@patrick-rodgers can you please review? I based the logic from the structure in pnp/sp.

### DIFF
--- a/debug/launch/graph.ts
+++ b/debug/launch/graph.ts
@@ -1,5 +1,6 @@
 import { Logger, LogLevel } from "../../packages/logging";
 import { graph } from "../../packages/graph";
+import "../../packages/graph/groups";
 import { AdalFetchClient } from "../../packages/nodejs";
 
 declare var process: { exit(code?: number): void };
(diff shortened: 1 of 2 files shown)


patrick-rodgers:
Going to merge this into 2.0.2 as I think it warrants it. This is likely a result of the rush to include graph immediately before launch - nice catch. I do think things were gettable using the invoke signature graph.me() for example - which is likely what I tested, but get needs to be included too for back compat. Thanks for submitting the fix!

PR #1055, opened 2020-02-06
Corrected tab's `displayName` property on v1

RoelVB:
#### Category
- [x] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?

Tab's name property should be `displayName` instead of `name`.
[See documentation](https://docs.microsoft.com/en-us/graph/api/teamstab-add?view=graph-rest-1.0)


patrick-rodgers:
Thanks @RoelVB - Appreciate the fix and pushing for both versions, that's a huge help!

Issue #1104, opened 2020-03-10
System.UriFormatException when using File.copyByPath()


mzbrandl:
### Category
- [x] Bug
- [x] Question

### Version

Please specify what version of the library you are using: [2.0.3]

Please specify what version(s) of SharePoint you are targeting: [online]

### Expected / Desired Behavior / Question
File.copyByPath() working as described in the [documentation](https://pnp.github.io/pnpjs/sp/files/#copy-by-path) and copying a file from one site collection to another.

### Observed Behavior
Following errors are thrown:
`POST https://xxx.sharepoint.com/sites/a/_api/SP.MoveCopyUtil.CopyFileByPath(overwrite=@a1)?@a1=false 500 (Internal Server Error)
Uncaught (in promise) Error: Error making HttpClient request in queryable [500] Internal Server Error ::> {"error":{"code":"-1, System.UriFormatException","message":{"lang":"en-US","value":"Invalid URI: The format of the URI could not be determined."}}}
at new HttpRequestError (parsers.js:130) 
at Function.<anonymous> (parsers.js:145) 
at step (tslib.es6.js:99)
at Object.next (tslib.es6.js:80)
at fulfilled (tslib.es6.js:70)`

### Steps to Reproduce
Create a SPFx web part, add pnp and deploy on a site. Get a file with getFileByServerRelativePath() works. Try to copy a file with .copyByPath() using either absolute url (https://xxx.sharepoint.com/sites/b/SitePages/test.aspx) or relative url (sites/b/SitePages/test.aspx) as parameter and providing either true or false as second parameter.

I hope this can be resolved, Thanks!

koltyakov:
`sites/b/SitePages/test.aspx` is incorrect relative URL. A relative URL must start with `/`. Could you try does it work for you with a correctly formatted relative URL (`/sites/b/SitePages/test.aspx`)?

mzbrandl:
I tried it with `/sites/b/SitePages/test.aspx`, same result. Then thats also an error for the destinationUrl in the code example in the [docs](https://pnp.github.io/pnpjs/sp/files/#copy-by-path) which is also missing `/`.

koltyakov:
Confirming the method doesn't work for me too, or I miss something (never used that method in my apps). @joelfmrodrigues looks that the method was added initially with your help, maybe you know what we miss? For me, it looks that it worked before and stopped due to possible server-side changes (e.g. expect only absolute paths for both `destPath` and `srcPath`),

[...]

PR #1108, opened 2020-03-10
Add additional checks for absolute Url

joelfmrodrigues:
#### Category
- [x] Bug fix?

#### Related Issues
mentioned in #1104

Add additional checks for absolute Url when copying or moving files/folders by path.
This is only to avoid errors if the user passes an absolute Url instead of relative.


patrick-rodgers:
Thanks for fixing this issue! Appreciate it!

PR #1113, opened 2020-03-17
Retry-After 429 throttling value to milliseconds, #1112

### LINKED ISSUE #1112: Retry-After 429 throttling value is used as milliseconds instead of seconds
### Category
- [ ] Enhancement
- [x] Bug
- [ ] Question
- [ ] Documentation gap/issue

### Version

Please specify what version of the library you are using: [1.3.7] but it exists in 1.3.10 and 2.0.3 codebase

Please specify what version(s) of SharePoint you are targeting: [SPO]

From what I could gather on the Net, the "Retry-After" value is specified in seconds, but the pnp code uses this value directly in the setTimeout method (milliseconds).

https://docs.microsoft.com/en-us/sharepoint/dev/general-development/how-to-avoid-getting-throttled-or-blocked-in-sharepoint-online

### Observed Behavior
On 429 throttling the "Retry-After" value is incorrectly used and the 7 retries are immediately executed without the expected delay.

sphttpclient.ts:
Line 88:
```
if (response.headers.has("Retry-After")) {
    // if we have gotten a header, use that value as the delay value
    delay = parseInt(response.headers.get("Retry-After"), 10);
}
```

Line 105:
```
// Set our retry timeout for {delay} milliseconds.
setTimeout(getCtxCallback(this, retry, ctx), delay);
```

### Steps to Reproduce
I used Fiddler to set the Response Statuscode and "Retry-After" Header to simulate throttling. Only after setting values above 1000 the delay was noticable.

Fiddler Script:
```
static function OnBeforeResponse(oSession: Session) {
    // SharePoint Throtteling Test
    if (oSession.url.indexOf("[mytenant].sharepoint.com") > -1
        && oSession.url.indexOf("/_api/web/getList") > -1
        && oSession.url.indexOf("GetClientSideComponents") === -1
    ) {
        oSession.responseCode = 429;
        oSession.ResponseHeaders["Retry-After"] = "3000";
    }
}
```

Thank you!

koltyakov:
#### Category

- [x] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?

#### Related Issues

fixes #1112

### DIFF
--- a/packages/sp/sphttpclient.ts
+++ b/packages/sp/sphttpclient.ts
@@ -86,8 +86,8 @@ export class SPHttpClient implements IRequestClient {
                 let delay: number;
 
                 if (response.headers.has("Retry-After")) {
-                    // if we have gotten a header, use that value as the delay value
-                    delay = parseInt(response.headers.get("Retry-After"), 10);
+                    // if we have gotten a header, use that value as the delay value in seconds
+                    delay = parseInt(response.headers.get("Retry-After"), 10) * 1000;
                 } else {
                     // grab our current delay
                     delay = ctx.delay;


patrick-rodgers:
Thanks for fixing that!

PR #1198, opened 2020-05-19
Removed type assertion and used the "as" syntax

derhallim:
#### Category
- [ ] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [x] Documentation update?

Type assertion doesn't work with .tsx files, so used the "as"  syntax instead that would work in ts and tsx files.


juliemturner:
Thanks. It makes sense, although we probably should be consistent then and make sure there aren't other instances.

PR #1216, opened 2020-06-05
Adds read operations for v2.1 Taxonomy API

### LINKED ISSUE #1160: A Note on Taxonomy Support in PnPjs 2.*
Hello,

Wanted to make a note as we have gotten this question a few times. We did not bring the old model taxonomy work into the 2.* branch of the library. This was a choice based on what we thought was the timeline for introduction of the new API. We are waiting, as I know are many of you, for an announcement and and API to test against - at which time we'll get support added into the 2.* branch ASAP.

This has taken longer than we hoped, but it was and remains the right call. It doesn't make sense to eventually support two sets of taxonomy operations within the same branch (IMO) as it would get confusing, they would work differently, and leaving behind the complexity of reverse engineering the jsom API into fluent is a good thing.

For now if you need to access the taxonomy using pnpjs you are stuck in the 1.* branch. That is not ideal, and we understand. But know we will add support into 2.* just as soon as it is available.

Thanks!

Edit: Should add this issue will be updated once we start work so folks have visibility into progress.

patrick-rodgers:
#### Category
- [ ] Bug fix?
- [X] New feature?
- [ ] New sample?
- [ ] Documentation update?

#### Related Issues

fixes #1160 

Adds the docs, method support, and tests for the v2.1 taxonomy read operations.

Update and Create are not included currently due to time and determining permissions/auth requirements. Will look to add in next version as possible.


TheShihan:
Any update on the write functions? I am on 2.0.7 and just found out that I am not able to add a term :-(

PR #1259, opened 2020-07-09
WIP: Add getStream

naugtur:
#### Category
- [ ] Bug fix?
- [x] New feature?
- [ ] New sample?
- [ ] Documentation update?

I had to put a lot of research in to figure out how to get a stream instead of loading the whole thing into memory.
Without reading the code of getBlob I wouldn't be able to do it.
This is a WIP to check if the contribution like this is welcome.

I'm aware the parser would need to go to the odata package and inherit properly.
I'm aware docs would follow.

It seemed odd this method is not there yet, so I'm checking if there wasn't a reason behind it.

Please comment.

### DIFF
--- a/packages/odata/parsers.ts
+++ b/packages/odata/parsers.ts
@@ -5,6 +5,11 @@ export interface IODataParser<T> {
     parse(r: Response): Promise<T>;
 }
 
+export interface IResponseBodyStream {
+    body: ReadableStream<Uint8Array>;
+    knownLength: number;
+}
+
 export class ODataParser<T = any> implements IODataParser<T> {
 
     public parse(r: Response): Promise<T> {
@@ -89,6 +94,12 @@ export class BlobParser extends ODataParser<Blob> {
     protected parseImpl(r: Response, resolve: (value: any) => void): void {
 
         r.blob().then(resolve);
+    }
+}
+export class StreamParser extends ODataParser<IResponseBodyStream> {
+
+    protected parseImpl(r: Response, resolve: (value: any) => void): void {
+        resolve({ body: r.body, knownLength: parseInt(r.headers['content-length'],10)})
     }
 }
 
--- a/packages/odata/parsers.ts
+++ b/packages/odata/parsers.ts
@@ -99,7 +99,7 @@ export class BlobParser extends ODataParser<Blob> {
 export class StreamParser extends ODataParser<IResponseBodyStream> {
 
     protected parseImpl(r: Response, resolve: (value: any) => void): void {
-        resolve({ body: r.body, knownLength: parseInt(r.headers['content-length'],10)})
+        resolve({ body: r.body, knownLength: parseInt(r.headers["content-length"], 10) });
     }
 }
 
--- a/packages/odata/queryable.ts
+++ b/packages/odata/queryable.ts
@@ -13,6 +13,13 @@ import { PipelineMethod } from "./pipeline";
 import { IODataParser, ODataParser } from "./parsers";
 
 export function cloneQueryableData(source: Partial<IQueryableData>): Partial<IQueryableData> {
+  let body;
+  // this handles bodies that cannot be JSON encoded (Blob, etc)
+  // Note however, even bodies that can be serialized will not be cloned.
+  if (source.options && source.options.body) {
+    body = source.options.body;
+    source.options.body = "-"
+  }
 
   const s = JSON.stringify(source, (key: string, value: any) => {
 
@@ -53,9 +60,8 @@ export function cloneQueryableData(source: Partial<IQueryableData>): Partial<IQu
     }
   });
 
-  // this handles bodies that cannot be JSON encoded (Blob, etc)
-  if (source.options && source.options.body) {
-    parsed.options.body = source.options.body;
+  if (body) {
+    parsed.options.body = body;
   }
 
   return parsed;
--- a/packages/odata/queryable.ts
+++ b/packages/odata/queryable.ts
@@ -18,7 +18,7 @@ export function cloneQueryableData(source: Partial<IQueryableData>): Partial<IQu
   // Note however, even bodies that can be serialized will not be cloned.
   if (source.options && source.options.body) {
     body = source.options.body;
-    source.options.body = "-"
+    source.options.body = "-";
   }
 
   const s = JSON.stringify(source, (key: string, value: any) => {
--- a/packages/odata/queryable.ts
+++ b/packages/odata/queryable.ts
@@ -62,6 +62,8 @@ export function cloneQueryableData(source: Partial<IQueryableData>): Partial<IQu
 
   if (body) {
     parsed.options.body = body;
+    // Since we're attempting to clone the source to avoid changing it, the least we can do with the body is keep the reference intact.
+    source.options.body = body;
   }
 
   return parsed;
--- a/packages/sp/files/types.ts
+++ b/packages/sp/files/types.ts
@@ -7,7 +7,7 @@ import {
     IDeleteableWithETag,
     deleteableWithETag,
 } from "../sharepointqueryable";
-import { TextParser, BlobParser, JSONParser, BufferParser, headers, body } from "@pnp/odata";
+import { TextParser, BlobParser, JSONParser, BufferParser, StreamParser, headers, body, IResponseBodyStream } from "@pnp/odata";
 import { assign, getGUID, isFunc, stringIsNullOrEmpty, isUrlAbsolute } from "@pnp/common";
 import { Item, IItem } from "../items";
 import { odataUrlFrom } from "../odata";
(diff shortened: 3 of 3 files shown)


naugtur:
BTW. Here's my current implementation - streams work fine, but documentation was not enough to arrive at this, I had to read code of the library:

```js
async function getSelectedFileAsStream(selectedFile) {
  // There are two things needed to get the file content
  // 1. adding to the URL:
  // selectedFile.concat("/$value")
  // the /$value bit is what needs to be added to URL to get the data back
  // 2. adding a magic header
  // .usingParser(Parser)({headers:{ "binaryStringResponseBody": "true" }})
  return await selectedFile.concat("/$value").usingParser({
    // custom parser implementation to avoid consuming the stream and expose it instead
    parse(r) {
      return Promise.resolve({ stream: r.body, headers: r.headers });
    },
  })({ headers: { binaryStringResponseBody: "true" } }); 
}
```

patrick-rodgers:
Apologies for the delay in responding, I have been out on family leave and am working to get caught up.

Thanks @naugtur - first PRs like this are always welcome and appreciate you taking the time. Just so I understand this would be a Node feature to support node streams? I think that is something we can def look at supporting and as you noted possibly combined with #1261.

I think the way to do this would be as you describe by adding a stream parser in the odata parsers module and updating the docs. If this will only work in node (I think so?) then we just need to call that out. OR something to consider if it is really node only should we add it to the node package?

To answer why we haven't done something like this it is simply a question of time and resources. The bulk of our users are browser based, so more of our investment of time goes into supporting those features. BUT we value node just as much and I know folks are using this library in node - so always excited to improve that experience.

naugtur:
Don't worry about the response time, you have no obligations as per the license :)

I didn't expect a response within the first week, but that won't stop me from sending stuff your way.

To the point, I intend to expand this PR. To cover all the needed items, so I could use some guidance.
I'll share my attempt here soon for review.

Using blobs/buffers in Node.js is very rarely a reasonable choice for files and streams are the best way to pass files through an app.

There's the related issue with the limited support for streams when uploading. It will need your intervention.

Speaking of where it'd work, streams are node only. The method would explicitly have to be documented as providing support for node streams.

The only compatibility layer on top of browser and node streams are async iterators but I feel it's a little too early to introduce support for that.

Anyway, I'm glad you accept my intention to provide support and let's get to work!

naugtur:
Update. @patrick-rodgers Please let me know if this is moving in the right direction.
I'm not using TypeScript on daily basis, so the way I defined things might not be right.

packages/odata/parsers.ts
```diff
@@ -5,6 +5,11 @@ export interface IODataParser<T> {
     parse(r: Response): Promise<T>;
 }
 
+export interface IResponseBodyStream {
+    body: ReadableStream<Uint8Array>;
```
naugtur, line 9: This is not perfect, but it matches the type I get as the request body.

packages/odata/parsers.ts
```diff
@@ -91,6 +96,12 @@ export class BlobParser extends ODataParser<Blob> {
         r.blob().then(resolve);
     }
 }
+export class StreamParser extends ODataParser<IResponseBodyStream> {
+
+    protected parseImpl(r: Response, resolve: (value: any) => void): void {
+        resolve({ body: r.body, knownLength: parseInt(r.headers["content-length"], 10) });
```
naugtur, line 102: FYI: I'd gladly return just the stream to make the interface cleaner, but in many cases a stream without known length is not useful for sending further.

packages/odata/queryable.ts
```diff
@@ -13,6 +13,13 @@ import { PipelineMethod } from "./pipeline";
 import { IODataParser, ODataParser } from "./parsers";
 
 export function cloneQueryableData(source: Partial<IQueryableData>): Partial<IQueryableData> {
+  let body;
+  // this handles bodies that cannot be JSON encoded (Blob, etc)
+  // Note however, even bodies that can be serialized will not be cloned.
```
naugtur, line 18: The original code would not attempt to clone the body either.
I didn't change that behavior. 

It was doing all the work to `JSON.stringify` it though and that's where it failed if passed Node's IncomingMessage type stream with all the response and network additions.

naugtur: This should address https://github.com/pnp/pnpjs/issues/1261

packages/sp/files/types.ts
```diff
@@ -46,7 +46,7 @@ export class _Files extends _SharePointQueryableCollection<IFileInfo[]> {
      * @returns The new File and the raw response.
      */
     @tag("fis.add")
-    public async add(url: string, content: string | ArrayBuffer | Blob, shouldOverWrite = true): Promise<IFileAddResult> {
+    public async add(url: string, content: string | ArrayBuffer | Blob | any, shouldOverWrite = true): Promise<IFileAddResult> {
```
naugtur, line 49: `any` is what fetch defines as its input anyway

packages/odata/queryable.ts
```diff
@@ -13,6 +13,13 @@ import { PipelineMethod } from "./pipeline";
 import { IODataParser, ODataParser } from "./parsers";
 
 export function cloneQueryableData(source: Partial<IQueryableData>): Partial<IQueryableData> {
```
naugtur, line 15: I couldn't find tests for this. 
I could write a test covering the cases with body but I'd need at least a stub with all the setup.

packages/odata/parsers.ts
naugtur: Is this where it belongs though?

naugtur:
Hi, @patrick-rodgers 
Any thoughts on the work in this PR?

Could you tell me if this is portable to the graph API part of the library? I might end up switching to graph for the sake of preferable auth setup.

patrick-rodgers:
Hey there! - I haven't forgotten about this, just been busy with work and getting the 2.0.7 release out. I haven't had a chance to go through all your comments but I have bene thinking on this and wonder if putting the node functionality into the nodejs package might be a better approach? Again, I haven't had a chance to pull this and have a good look yet.

patrick-rodgers:
OK, forced myself to sit down and review this since it has been sitting too long with a response, apologies for that.

- Based on [this article](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API/Using_readable_streams) there is some browser support so leaving the code where it is makes sense to me
  - BUT I don't think the code is identical in browser or node so maybe this is node only??
  - Seem to be different stream APIs based on my reading?
- I am not sure what you're trying to do with your change in the clone method with the body vars. Doesn't seem needed?
- For file add we can just change the param type to any, no reason to list types and append any as it no longer is restrictive IMO
- How are testing this? It is typed as a ReadableStream which is a browser type. In my Node (LTS 10) the body comes through as a PassThrough so the typings aren't matching up - is it different in later versions? I had to do the below to test:

```ts
  const w = await sp.web.getFileByServerRelativePath("/sites/dev/Shared Documents/profile.jpg").getStream();

  // retype it to work in node
  const reader: PassThrough = <any>w.body;

  await new Promise((resolve) => {

    reader.on("data", (chunk) => {
      console.log(`Received ${chunk.length} bytes of data.`);
    });
    reader.on("end", () => {
      resolve();
    });
  });
```

naugtur:
Sorry my response is delayed - I postponed it because I had trouble finding out what you're referring to. 
Moving stuff to inline comments.

naugtur:
>     * Based on [this article](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API/Using_readable_streams) there is some browser support so leaving the code where it is makes sense to me
>       
>       * BUT I don't think the code is identical in browser or node so maybe this is node only??
>       * Seem to be different stream APIs based on my reading?

Node had to invent streams long before any standard was on the horizon. This is adding support for Node.js Streams as these are the only ones currently available. Browser streams are slowly rolling out, but they're not compatible with node streams and are likely to remain incompatible in the future. The compatibility layer between those (and anything stream-alike for that matter) is async iterators. But most node devs need the streams.

TL;DR: this is Node.js only.

packages/sp/files/types.ts
naugtur: @patrick-rodgers 
> For file add we can just change the param type to any, no reason to list types and append any as it no longer is restrictive IMO

We can collapse this to `any` or add some types representing node streams (would have to investigate if there's a single root type for all streams) but this value is passed to fetch, which accepts `any`.

Keeping the other types has one advantage - IDE suggestions. "some type hints or any" is better than just "any". 

I'd keep the list as long as possible for the sake of good IDE tips.

naugtur:
> How are testing this? It is typed as a ReadableStream which is a browser type. In my Node (LTS 10) the body comes through as a PassThrough so the typings aren't matching up - is it different in later versions? I had to do the below to test:

I tested it by using it in a PoC appllication. 
Well, I've only tested it in JS on a modified built version. The code is sound, the types have been made up when writing the PR. That's where I need your help I guess - to type it reasonably.

As for the returned stream - PassThrough is also a Readable Stream, at least conceptually.

PR #1264, opened 2020-07-13
sphttpclient: added IFetchOptions parameter for digest retrieval logic

f1nzer:
#### Category
- [x] Bug fix

Digest retrieval logic (/api/contextinfo) now ignores IFetchOptions from the request.
The problem is that I have custom `credentials` value (`includes`) that is ignored during the request (`same-origin` is used all the time), so 401 error appears.

### DIFF
--- a/packages/sp/sphttpclient.ts
+++ b/packages/sp/sphttpclient.ts
@@ -59,7 +59,7 @@ export class SPHttpClient implements IRequestClient {
         // if we have either a request digest or an authorization header we don't need a digest
         if (opts.method && opts.method.toUpperCase() !== "GET" && !headers.has("X-RequestDigest") && !headers.has("Authorization")) {
 
-            const digest = await this._digestCache(extractWebUrl(url));
+            const digest = await this._digestCache(extractWebUrl(url), opts);
             headers.append("X-RequestDigest", digest);
         }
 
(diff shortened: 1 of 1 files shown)


patrick-rodgers:
Hi @f1nzer - I am reluctant to make this change as it is deep in the core of the library. Have a look at [this line](https://github.com/pnp/pnpjs/blob/version-2/packages/sp/sphttpclient.ts#L60) where we are skipping digest if we already have an authorization header. Is there a similar thing we can use to determine your auth doesn't need the digest? Do you still require teh digest if your auth is present? Without knowing the details of whatever custom authentication is in play its hard to give suggestions but would want more details before merging this. Thanks!

f1nzer:
Hi @patrick-rodgers 
What I am doing is a SPA outside of SharePoint (inside an intranet). For authorization I'm using `credentials: 'include'` - that works in general (digest is not needed for basic REST queries), but for CAML queries I have to use the digest from `/api/contextinfo` and this query should contain `credentials: 'include'` setting (but `same-origin` is strictly used now).

As a workaround, I have to use a global window.fetch interceptor, but that's really hacky.

juliemturner:
Although I appreciate your special circumstances, your update, although it may work for you, breaks how the library works today, and would require everyone to build their own options. 

What @patrick-rodgers and I have discussed is the potential making some small modifications that would allow subclassing of the SPHttpClient so that you can build your own implementation. Going this route then allows anyone who has special circumstances to manage their own work arounds but doesn't fundamentally change the library. We'll look at making the modifications necessary in 2.0.8.

I'm going to create a separate issue to track this medication but will point to this PR that I'm going to now close.  We appreciate you sharing your time to help improve the library and hope this work around works for you.

PR #1267, opened 2020-07-13
fixing issue #1208, add uri encoding for filter, select, expand, and …

### LINKED ISSUE #1208: Special characters (& = +) URL encoding in .filter() function
### Category
- [ ] Enhancement
- [X] Bug
- [ ] Question
- [ ] Documentation gap/issue

### Version
Please specify what version of the library you are using: [ 2.0.5 ]

Please specify what version(s) of SharePoint you are targeting: [ Online ]

A file or page may have the name "Test & = + Test.aspx" in SharePoint. This may not be a good idea, but is allowed. If querying the list with PnP.js I should be able to use a filter like this:
```
import { sp } from "@pnp/sp/presets/all";
(async () => { 
  const listItems = await sp.web.lists.getByTitle("Site Pages")
			.items
			.filter(`FileRef eq '/sites/example/SitePages/Test & = + Test.aspx'`)
			.get();
  console.log("Listitems found: ", listItems.length);
})().catch(console.log)
``` 

### Observed Behavior
This results in a query that looks like:
https://contoso.sharepoint.com/sites/example/_api/web/lists/getByTitle('Site%20Pages')/items?$filter=FileRef%20eq%20%27/sites/example/SitePages/Test%20&%20=%20+%20Test.aspx%27

This results in a HTTP 400 InvalidQueryException `"message":{"lang":"de-DE","value":"Der Ausdruck \"FileRef eq '/sites/example/SitePages/Test \" ist ung\u00fcltig."}}}` because the & indicates the end of the query parameter.
The problematic characters & + and = are not urlencoded and may lead to query errors. Other special characters like spaces and umlauts are urlencoded correctly. 

I tried to perform url encoding on my own like this:
```
...
.filter(`FileRef eq '${encodeURIComponent('/sites/example/SitePages/Test & = + Test.aspx')}'`)
...
```
But then characters like spaces are encoded twice: %2520 instead of %20. That also leads to errors.

It looks to me that in pnpjs the encodeURI() function is used instead of encodeURIComponent(). Unfortunately I didn't find the code piece where this happens to dig deeper.

### Steps to Reproduce
Try out the above example code.

Thank you for your support!
Tobias

patrick-rodgers:
…order by query params

#### Category
- [X] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?

#### Related Issues

fixes #1208

Adds encoding for filter, select, expand, and order by query params when they are added through the methods. Change applies to both sp and graph

### DIFF
--- a/packages/graph/graphqueryable.ts
+++ b/packages/graph/graphqueryable.ts
@@ -57,7 +57,7 @@ export class _GraphQueryable<GetType = any> extends Queryable<GetType> implement
      */
     public select(...selects: string[]): this {
         if (selects.length > 0) {
-            this.query.set("$select", selects.join(","));
+            this.query.set("$select", selects.map(encodeURIComponent).join(","));
         }
         return this;
     }
@@ -69,7 +69,7 @@ export class _GraphQueryable<GetType = any> extends Queryable<GetType> implement
      */
     public expand(...expands: string[]): this {
         if (expands.length > 0) {
-            this.query.set("$expand", expands.join(","));
+            this.query.set("$expand", expands.map(encodeURIComponent).join(","));
         }
         return this;
     }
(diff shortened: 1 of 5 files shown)


johannes-z:
@patrick-rodgers I think this also applies to `RenderListDataAsStream`'s `ViewXml`.

PR #1272, opened 2020-07-14
mkdocs: remove duplicate Search item from the docs menu

f1nzer:
#### Category
- [x] Documentation update

"Search" item is duplicated in the docs menu, so one can be removed.
image


patrick-rodgers:
Nice catch, thanks for taking the time to submit the PR to fix it!

PR #1306, opened 2020-07-31
feat(list): get validation message endpoint

JakeStanger:
#### Category
- [ ] Bug fix?
- [x] New feature?
- [ ] New sample?
- [ ] Documentation update?

Adds a missing mapping for an endpoint:

```ts
sp.web.lists.getByTitle("Some List").validationMessage().then(console.log) // → validation message string
```

Note: this is the user message set in the list validation settings.

### DIFF
--- a/packages/sp/lists/types.ts
+++ b/packages/sp/lists/types.ts
@@ -172,7 +172,7 @@ export class _List extends _SharePointQueryableInstance<IListInfo> {
      * Gets the user validation message for this list
      *
      */
-    public get validationMessage() {
+    public get validationMessage(): ISharePointQueryable<string> {
         return tag.configure(SharePointQueryable(this, "validationMessage"), "l.validationMessage");
     }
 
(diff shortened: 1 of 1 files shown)


patrick-rodgers:
Hi @JakeStanger -

Thanks for this, curious why this doesn't work for you?

```ts
const message = await sp.web.lists.getByTitle("My Title").select("ValidationMessage")();
```

We don't typically add a property for every object property that exists so long as it is selectable. This always works too:

```ts
  // you do have to type it as any because ValidationMessage isn't in the interface, fixing that now
  const w = await sp.web.lists.getByTitle("My Title").update(<any>{
    ValidationMessage: "New Message",
  });
```

JakeStanger:
The only documentation I could find on how to get the validation message suggested you had to use a `/_api/Web/Lists(guid'')/validationMessage` endpoint. I'll be honest, I tried it and it worked, and I never thought to try just using `$select` nor was I aware there's (at least some) interchangeability between the two formats.

Given that just using `$select` on the list works fine i think we're fine to close & forget this.

patrick-rodgers:
Right on. Almost all the properties can also be used in the web/lists/get/property format. There are a small few that you have to access that way. There is also the issue where not all properties that are available are returned if you don't select for them.

Do appreciate you taking the time to create the PR, even if in this case it isn't needed. Thanks!!

PR #1345, opened 2020-08-26
fix for #1311

### LINKED ISSUE #1311: Calling `getParentWeb` on the root Web results in a call to `_api/site/openWebById('undefined')`
### Category
- [ ] Enhancement
- [X] Bug
- [ ] Question
- [ ] Documentation gap/issue

### Version

Please specify what version of the library you are using: [2.0.6]

Please specify what version(s) of SharePoint you are targeting: [2016]

I created a Web that points to the root web of an instance, and thought if I tried to call `getParentWeb` that it would fail gracefully in some way.  (Maybe there's a better way to tell that my Web instance is the site root?)

### Observed Behavior
It actually makes a request to `_api/site/openWebById('undefined')` (with the string literal "undefined" there) and bubbles out the network error that results.

### Steps to Reproduce
````ts
const web = Web(url);
const parent = await web.getParentWeb();   // throws
````

----

As an aside, if anybody has a better suggestion for how to check that a given Web represents the site root, I'd be happy to hear it.

patrick-rodgers:
#### Category
- [X] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?

#### Related Issues

fixes #1311

Fixes an issue for users not using minimal meta-data where unpatched 2016 returns verbose results that caused an error in getParentWeb.

### DIFF
--- a/packages/sp/webs/types.ts
+++ b/packages/sp/webs/types.ts
@@ -85,7 +85,7 @@ export class _Web extends _SharePointQueryableInstance<IWebInfo> {
     @tag("w.getParentWeb")
     public async getParentWeb(): Promise<IOpenWebByIdResult> {
         const { ParentWeb } = await spGet(this.select("ParentWeb/Id").expand("ParentWeb"));
-        return ParentWeb ? Site(this.parentUrl).openWebById(ParentWeb.Id) : null;
+        return ParentWeb?.Id ? Site(this.parentUrl).openWebById(ParentWeb.Id) : null;
     }
 
     /**


thw0rted:
Thanks, Patrick.  When you say "unpatched 2016", is there a fix that's supposed to cause a different response format?  I thought this instance was up to date.

thw0rted:
Hi @patrick-rodgers , I just had a chance to test this as provided in v2.0.10.  It does prevent the invalid fetch (with a literal `undefined` in the URL), but it is missing one thing.  The promise returned by `getParentWeb()` is typed as resolving a `IOpenWebByIdResult` but when called on the root web, the it actually resolves with `null`.  The typings should be updated to reflect this.

PR #1393, opened 2020-09-28
Use string values for the BasePermission High and Low properties in SP RoleDef updates.

danwatford:
#### Category
- [X] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?

Code change and supporting test to allow updating an SP RoleDef's BasePermission value.

Current implementation passes numbers rather than strings for the BasePermission Low and High value when updating a RoleDef. Strings are already used for these properties when creating a RoleDef so this PR does the same for updates.

See https://docs.microsoft.com/en-us/previous-versions/office/developer/sharepoint-rest-reference/dn531432(v=office.15)#roledefinition-resource for use of strings in REST calls updating a RoleDefinition resource.

Please note that the test operates on the parent web, rather than the temporary subsite created during a test run. This is due to RoleDefs being inherited by subsites meaning any RoleDef testing must be performed on a root site.

### DIFF
--- a/packages/sp/security/types.ts
+++ b/packages/sp/security/types.ts
@@ -170,6 +170,10 @@ export class _RoleDefinition extends _SharePointQueryableInstance<IRoleDefinitio
         const s = ["BasePermissions"];
         if (hOP(properties, s[0]) !== undefined) {
             properties[s[0]] = assign(metadata(`SP.${s[0]}`), properties[s[0]]);
+
+            const bpObj = properties[s[0]];
+            bpObj.High = bpObj.High.toString();
+            bpObj.Low = bpObj.Low.toString();
         }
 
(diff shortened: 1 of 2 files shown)


danwatford:
Without this fix, attempts to update the BasePermissions of a RoleDefinition result in the following response from SP:

> Error: Error making HttpClient request in queryable [400] Bad Request ::> {"odata.error":{"code":"-1, Microsoft.SharePoint.Client.InvalidClientQueryException","message":{"lang":"en-US","value":"Cannot convert a primitive value to the expected type \'Edm.Int64\'. See the inner exception for more details."}}}

patrick-rodgers:
Thanks, will have a look at getting this merged for the next release. Appreciate you submitting it!

PR #1402, opened 2020-10-07
#1395 removed 'odata.id' dependency

### LINKED ISSUE #1395: Method Folder.moveByPath: "Invalid URI: The format of the URI could not be determined." -> odata.id missing
Hello

I am using the method moveByPath to move a folder within a doc library.
I am using version 2.0.8

When I call the method I get the error message ""Invalid URI: The format of the URI could not be determined."
I compared my request, with a request that is done within SPO when I move a folder using the UI. These requests are using absolute URLs. My request the URLs are **relative**. 

I have checked the code of the moveByPath function and the API Request to _api/SP.MoveCopyUtil.MoveFolderByPath() should always use **absolute** paths.

Within debugging I recogniced that the code is expacting an "odata.id" object from the Folder.Select(...) request. This "odata.id" is not within the response, so in line 185 of the file "https://github.com/pnp/pnpjs/blob/version-2/packages/sp/folders/types.ts" you are expecting the "odata.id" to containing the absolute url.

As this object is not present the hosturl is build empty and the request is relative.

I assume that these relative URLs are causing the "Invalid URI" exception...

Thanks for checking and your help
Pascal

moveByPath-Request:
image

Folder.Select(...) Request within moveByPath Method
image

hohenp:
#### Category
- [x] Enhancement

#### Related Issues
fixes #1395 

The methods (moveTo, moveByPath, copyTo, copyByPath) had an dependency to the "odata.id" which is not present using odata:verbose configuration. The "odata.id" was used to extract the absolute URL. I am now using the Url and ServerRelativeUrl of the web object to get the absolute URL and the host Url.

### DIFF
--- a/packages/sp/folders/types.ts
+++ b/packages/sp/folders/types.ts
@@ -21,6 +21,7 @@ import { escapeQueryStrValue } from "../utils/escapeQueryStrValue";
 import { extractWebUrl } from "../utils/extractweburl";
 import { tag } from "../telemetry";
 import { toResourcePath, IResourcePath } from "../utils/toResourcePath";
+import { sp } from "../rest";
 
 @defaultPath("folders")
 export class _Folders extends _SharePointQueryableCollection<IFolderInfo[]> {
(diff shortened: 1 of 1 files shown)


packages/sp/folders/types.ts
```diff
@@ -162,10 +163,10 @@ export class _Folder extends _SharePointQueryableInstance<IFolderInfo> {
     @tag("f.moveTo")
     public async moveTo(destUrl: string): Promise<void> {
 
-        const { ServerRelativeUrl: srcUrl, ["odata.id"]: absoluteUrl } = await this.select("ServerRelativeUrl")();
-        const webBaseUrl = extractWebUrl(absoluteUrl);
-        const hostUrl = webBaseUrl.replace("://", "___").split("/")[0].replace("___", "://");
-        await spPost(Folder(webBaseUrl, "/_api/SP.MoveCopyUtil.MoveFolder()"),
+        const { ServerRelativeUrl: srcUrl } = await this.select("ServerRelativeUrl")();
+        const web = await sp.web.select("Url, ServerRelativeUrl")();
```
koltyakov, line 167: This won't work, unfortunately. `sp.web` takes the current page web or a baseUrl-provided web. However, `this` in the class instance is not necessarily chained from a current web, but any defined for a parent API instance.

koltyakov:
Thanks for the PR. The general comment about internals: `sp.web` shouldn't be used inside methods as it does not always end up with a web to which the current chainable object belongs to.

I guess that we can extract web URL using ```const webUrl = extractWebUrl(this.toUrl());```. Can check it later on, always in a rush these weeks.

hohenp:
Hi Andrew, thanks for your help and review. I didn't really felt comfortable using sp.web but didn't see another way to get the absolute url. I thought as the "/_api/SP.MoveCopyUtil..." API is getting the srcPath and destPath it might be not that crucial, if the web url is not reflected from "this".

I tried this.ToUrl() while debugging, but got:
image

a context object or this.ParentWeb does not exist either...

patrick-rodgers:
Thanks! We will have a look and get it into the next release! Appreciate you submitting this 😀

patrick-rodgers:
Thanks @hohenp! We ended up rewriting it a bit to include new methods for getParentInfos for List, Item, and Folder to make this kind of operation easier.

Issue #1513, opened 2021-01-06
How to upload bigger than 2GB file to Share Point Online with PnPjs?


ntodorov:
### Category
- [x] Enhancement
- [ ] Bug
- [x] Question
- [ ] Documentation gap/issue

### Version

Please specify what version of the library you are using: [    2.0.13     ]

Please specify what version(s) of SharePoint you are targeting: [     SPO    ]

### Expected / Desired Behavior / Question

The man question: **How to upload >2GB file to SPO without loading it in memory with fs.readFile()?**

I do not see a good example of uploading very large local files with pnpjs under nodejs.
With the help of @koltyakov I managed to adapt the example of using nodejs streams shown here:
https://pnp.github.io/pnpjs/sp/files/#adding-a-file-using-nodejs-streams
to use `fs.createReadStream()`, but I hit the SharePoint Online limitation -> File size (request body) is limited to 262144000 bytes (more info: https://github.com/koltyakov/pnp-upload/issues/6). 
Unfortunately, that's 262MB so not really workable for my 2GB and bigger files.

Up to 2 GB can be uploaded with `addChunks()` but that requires loading in memory first in order to be chunked.

patrick-rodgers:
Curious what you think we can do if the service has a 2 GB limit?

koltyakov:
Nik is referencing slightly bit different thing. 

File size limit for a file in SPO library is greater than 2 GB (10 or 100 GB). Yet, with ordinary file.add even providing a readable stream to the method the limit is 250 MB, as this is a max size per a request body. Here, we can do nothing.

With addChunks, it’s possible uploading up to the limit of 10/100 GB (a brave assumption), but... addChunks uses Blob/ArrayBuffer therefore to upload a file its whole body should be loaded to the memory first then passes to the method. With a readable stream, next chunk read could happening in a stream version of addChunks (which I could create and PR) before calling continueUpload/other internal methods used in addChunks. So the memory consumption won’t be sized to a file volume but a single chunk.

koltyakov:
Before creating a stream version of addChunks, the open question for a brief research: maybe there is any other simple way uploading file as stream and not facing body length limit. E.g. providing content type header (application/octet-stream;) or something.

jansenbe:
[...]

PR #1517, opened 2021-01-10
addChunked stream version method for Node.js, #1513

koltyakov:
#### Category

- [ ] Bug fix?
- [x] New feature?
- [ ] New sample?
- [ ] Documentation update?

#### Related Issues

mentioned in #1513

The PR adds `addChunked` method extension to support ReadStream in Node.js.

#### Guidance

For a test, run `npm run test -- -g 'nodejs - sp-extensions'`

Usage: when Node.js `sp-extensions` are loaded `addChunked` method is extended and "polyfilled" to support stream content received with Node's `fs.createReadStream`.

```typescript
const stream = fs.createReadStream('/path/to/file.dat');
await files.addChunked(name, stream);
```

Existing user implementations should not fail. When no-stream content is passed, it is processed as previously.

### DIFF
--- a/packages/nodejs/sp-extensions/stream.ts
+++ b/packages/nodejs/sp-extensions/stream.ts
@@ -1,5 +1,10 @@
+import { getGUID, isFunc } from "@pnp/common/util";
 import { ODataParser, extendFactory, headers } from "@pnp/odata";
-import { File } from "@pnp/sp/files";
+import { File, Files, IFileAddResult, IFileInfo, IFileUploadProgressData } from "@pnp/sp/files";
+import { odataUrlFrom } from "@pnp/sp/odata";
+import { spPost } from "@pnp/sp/operations";
+import { escapeQueryStrValue } from "@pnp/sp/utils/escapeQueryStrValue";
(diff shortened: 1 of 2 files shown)


patrick-rodgers:
Awesome, thanks! We'll have a look and get it merged ahead of the release Friday so will land in 2.1.0.

patrick-rodgers:
TIL about the -g option for mocha. That is awesome. Thanks for doing this, will be a big help for folks in node. I added the methods to the docs page too.

koltyakov:
Oh, docs, that's what I forgot completely.

Tested the published version's method by uploading 3GB file and checking if I can playback it after. Went well. But found a nuance which we could add to docs as well. Readable stream should be created with fixed hight water mark equal to chunk size:

```typescript
// Important: highWaterMark must be equal to chunkSize
const rs = fs.createReadStream(filePath, { highWaterMark: chunkSize });
```

Linking a comment with usage sample for those who found a  the PR directly https://github.com/pnp/pnpjs/issues/1513#issuecomment-761147133

PR #1716, opened 2021-04-21
made sp.site.getContextInfo() compatible with SP2013

lukju:
#### Category
- [x ] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?

#### Related Issues
-

sp.site.getContextInfo() does not work on SP2013 as there is not way to set the 'Accept' HTTP header to 'application/json; odata=verbose'.
tried to configure sp.site this way:
sp.site.configure({headers: { 'accept': 'application/json; odata=verbose' } }).getContextInfo()
which does not work either.
My change ensures that the optional IFetchOptions object provided in the above configure() function call gets added to the getContextInfo() call...

### DIFF
--- a/packages/sp/sites/types.ts
+++ b/packages/sp/sites/types.ts
@@ -66,7 +66,7 @@ export class _Site extends _SharePointQueryableInstance {
     public async getContextInfo(): Promise<IContextInfo> {
 
         const q = tag.configure(Site(this.parentUrl, "_api/contextinfo"), "si.getContextInfo");
-        const data = await spPost(q);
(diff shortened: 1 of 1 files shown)


juliemturner:
Thanks we'll have a look.

patrick-rodgers:
Hi @lukju - you have indeed found a bug (thank you!!!) but I am going to update the fix slightly. The proper way to handle this would be to apply the configureFrom method, but there is no way for you to know that so no worries :). Very much appreciate your submitting the fix for this, thank you!!

PR #1774, opened 2021-05-25
Check if CorporateCatalogUrl before returning Web

allanhvam:
#### Category
- [X] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?

On tenants without an corp. app catalog getTenantAppCatalogWeb throws an error because data.CorporateCatalogUrl isn't set. Data returned from _api/SP_TenantSettings_Current is similar to:

<?xml version="1.0" encoding="utf-8"?><entry xml:base="https://tenant.sharepoint.com/_api/" xmlns="http://www.w3.org/2005/Atom" xmlns:d="http://schemas.microsoft.com/ado/2007/08/dataservices" xmlns:m="http://schemas.microsoft.com/ado/2007/08/dataservices/metadata" xmlns:georss="http://www.georss.org/georss" xmlns:gml="http://www.opengis.net/gml"><id>https://tenant.sharepoint.com/_api/SP_TenantSettings_Current</id><category term="SP.TenantSettings" scheme="http://schemas.microsoft.com/ado/2007/08/dataservices/scheme" /><link rel="edit" href="SP_TenantSettings_Current" /><title /><updated>2021-05-21T13:32:08Z</updated><author><name /></author><content type="application/xml"><m:properties><d:CorporateCatalogUrl m:null="true" /></m:properties></content></entry>

Note CorporateCatalogUrl m:null="true". This PR will check if CorporateCatalogUrl is set before constructing the Web.

### DIFF
--- a/packages/sp/appcatalog/index.ts
+++ b/packages/sp/appcatalog/index.ts
@@ -22,6 +22,9 @@ SPRest.prototype.getTenantAppCatalogWeb = async function (this: SPRest): Promise
 
     return this.childConfigHook(async ({ options, runtime }) => {
         const data: { CorporateCatalogUrl: string } = await SharePointQueryable("/", "_api/SP_TenantSettings_Current").configure(options).setRuntime(runtime)();
+        if (!data?.CorporateCatalogUrl) {
+            return undefined;
+        }
         return Web(data.CorporateCatalogUrl).configure(options).setRuntime(runtime);
     });
 };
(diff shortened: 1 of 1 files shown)


patrick-rodgers:
@allanhvam, nice catch here. Thanks!

Thinking this through - would it be better to throw an Error saying that no catalog exists? Not sure, I ask because I am thinking if we give someone back undefined in a place where they expect a web they will get an error that is unclear on cause. One thought would be they can test to see if they get the web back (something like if (web) {}) but many folks won't. Could update the docs to explain that calling pattern as another option.

What do you think?

P.S. your actual code looks fine, just thinking through the change and what would be best.

allanhvam:
Thanks for the feedback. Think an error would be the better solution, I updated the PR.

patrick-rodgers:
Great stuff, thank you! Will be part of 2.6.0 going out today.

Issue #1807, opened 2021-06-18
[Graph] Support for $search parameter


JakeStanger:
### Category
- [x] Enhancement
- [ ] Bug
- [ ] Question
- [ ] Documentation gap/issue

### Version

Please specify what version of the library you are using: [2.6.0]

Please specify what version(s) of SharePoint you are targeting: [SPO]

### Expected / Desired Behavior / Question

Documentation link: https://docs.microsoft.com/en-us/graph/query-parameters#search-parameter

I've just discovered Graph has support for a `$search` parameter which works on messages, people and directory objects (users/groups). I can immediately see use-cases for this so it'd be great to get it in here!

I'd be happy to take this on as a PR (or at least as much as I can) if needed/wanted. One sticking point:

Should the `.search` method exist regardless of endpoint being used and return errors where used incorrectly, or should it only appear where support exists? (I'm not sure on how possible this is in the library currently)

patrick-rodgers:
We can have a look at adding it, would only want it to show up where it works vs. on all objects leading people to errors. If you want to have a swing at it, please do - always appreciate your help!

JakeStanger:
I'm pretty swamped in work this week so I likely won't be able to have a look until next week, but if nobody's picked it up by then I'll have a play :+1:

PR #1818, opened 2021-07-01
feat(graph): implement search param

JakeStanger:
#### Category
- [ ] Bug fix?
- [x] New feature?
- [ ] New sample?
- [ ] Documentation update?

#### Related Issues

Implements #1807

Adds the `$search` query param for messages, people and directory objects. The typings were already present but not used.

```ts
import { graph } from '@pnp/graph/presets/all';

graph.users.search("displayName:Jake").get();
```
Documented here <https://docs.microsoft.com/en-us/graph/search-query-parameter#using-search-on-directory-object-collections>

I couldn't find any docs or tests around the existing params; if you'd like those added for this please let me know where they should be and I'll get that done :)

### DIFF
--- a/packages/graph/graphqueryable.ts
+++ b/packages/graph/graphqueryable.ts
@@ -261,7 +261,7 @@ export interface IGraphQueryableCollection<GetType = any[]> extends IInvokable, 
 }
 export const GraphQueryableCollection = graphInvokableFactory<IGraphQueryableCollection>(_GraphQueryableCollection);
 
-export class _GraphQueryableSearchableCollection extends _GraphQueryableCollection {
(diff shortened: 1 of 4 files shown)


juliemturner:
@JakeStanger - Thanks for much for this, you're work looks great but what I realized in testing it out is that it requires the ConsistencyLevel header set to eventual. This is fine but @patrick-rodgers and I feel like we should give that some due consideration to figure out the best way to handle it for people. So, we're moving it out to next month to give us some time to review.

JakeStanger:
Ah whoops, not sure how I missed that when testing it. Let me know if you want anything further from myself for it.

PR #1865, opened 2021-08-29
A Prettier ConsoleListener

thechriskent:
#### Category
- [ ] Bug fix?
- [x] New feature?
- [ ] New sample?
- [x] Documentation update?

#### Related Issues

None

- ConsoleListener Changes
  - Prefix option - Prepends a string to every console message (helpful for filtering)
  - Text color options - Text color can be customized using CSS color values and can be set for all messages or even individual logging levels
  - `Message: ` is only prepended when Data is present in the Log entry resulting in cleaner log messages
    - Previously `Logger.write("Hello");`  would see a console message of `Message: Hello`, now it will simply be `Hello`
    - When Data is specified it will use the same behavior as before. `Logger.log(data: {text: "wowee"}, level: LogLevel.Info, message: "Hello"});` will still see a console message of `Message: Hello Data: {"text":"wowee"}`
- Documentation Updates
  - logging/index.md has been updated with instructions on using the new options

I have *not* included any tests since there were no previous tests for ConsoleListener and implementing them will require us to capture standard output and compare it likely using an additional npm package.

I've also targeted v3 as requested by Patrick.


juliemturner:
What a lovely idea... thanks @thechriskent

patrick-rodgers:
This is brilliant, thanks @thechriskent!

PR #1868, opened 2021-08-30
fetch signal implementation

SuperioOne:
#### Category
- [ ] Bug fix?
- [x] New feature?
- [ ] New sample?
- [ ] Documentation update?

I added fetch abort [signal](https://developer.mozilla.org/en-US/docs/Web/API/AbortController/signal) support at the IFetchOptions. The pull request also contains some minor lint fixes.

My initial attempt was creating a custom fetch client and extend IFetchOptions from outside. Unfortunately, cloneQueryableData function creates a deep copy via Json serialization which breaks signal property before options hits the any custom/default client.

### DIFF
--- a/packages/common/net.ts
+++ b/packages/common/net.ts
@@ -12,6 +12,7 @@ export interface IConfigOptions {
 export interface IFetchOptions extends IConfigOptions {
     method?: string;
     body?: any;
+    signal?: AbortSignal;
 }
 
 export interface IHttpClientImpl {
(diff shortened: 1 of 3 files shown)


juliemturner:
@patrick-rodgers -- My initial instinct would be to move this to a behavior in v3, similar to our advanced caching scenarios that override execute on the timeline?

patrick-rodgers:
@SuperioOne thank you for this contribution. I agree with Julie, I'd love to bring this change into v3. Can we work with you to move it to the version-3 branch?

If you have an immediate need for this we can likely also merge this to v2, but we're trying to push new work into v3 where possible.

SuperioOne:
@patrick-rodgers It's not a blocker for me so it can be pushed to v3. I'll try to add it as behaviour for new TimelinePipes.

PR #1936, opened 2021-11-08
added getUsingAbsoluteUrl method and updated files.add to use new method

### LINKED ISSUE #1914: Calling file.getItem() directly after uploading the file causes a timeout
### Category
- [ ] Enhancement
- [x] Bug
- [ ] Question
- [ ] Documentation gap/issue

### Version

Please specify what version of the library you are using: [    2.9.0     ]

Please specify what version(s) of SharePoint you are targeting: [  online   ]

Uploading a file to sharepoint will result in receiving a file object. Calling file.getItem() on this directly after uploading the file should work so that the file ID on sharepoint can be checked. 

### Observed Behavior
Calling file.getItem() directly after uploading the file will cause a timeout. 

### Steps to Reproduce
Code below always results in a timeout on file.getItem()
```
  const { file } = await sp.web
      .getFolderByServerRelativeUrl('/sites/krystian-test/Shared Documents/Test')
      .files.configure({ headers: { 'Content-Length': `${size}` } })
      .add('pug.jpg', stream, true);

      const fileMetadata = await file.getItem();
```

In my testing this always resulted in a timeout. 

On the other hand if another method is used to access file data it will work. For comparison the code below called right after uploading works fine.
```
  const fileMetadata = await sp.web
    .getFileByServerRelativeUrl(
      '/sites/krystian-test/Shared Documents/Test/pug.jpg'
    )
    .getItem();
```

aaademosu:
#### Category
- [x] Bug fix?
- [x] New feature?
- [ ] New sample?
- [ ] Documentation update?

#### Related Issues

fixes #1914

*This fixes the issue with calling file.getItem() in the ffg scenarios:*
- When the content uploaded is a stream using files.add rather than files.addChunked
- When the document library has over 5000 items which exceeds the SP online threshold limit.

The change made leverages the odata.id returned after an upload which is the absolute url of the file rather than use getFolderByServerRelativeUrl(*relativepath*)/files(*filename*)/listItemAllFields

### DIFF
--- a/packages/sp/files/types.ts
+++ b/packages/sp/files/types.ts
@@ -38,6 +38,15 @@ export class _Files extends _SharePointQueryableCollection<IFileInfo[]> {
     }
 
     /**
+     * Gets a File by file absolute url
+     *
+     * @param fileUrl The absolute url of the file
+     */
+     public getUsingAbsoluteUrl(fileUrl: string): IFile {
+        return tag.configure(File(`${fileUrl}`), "fis.getFileUsingResponseUrl");
+    }
+
+    /**
      * Uploads a file. Not supported for batching
      *
      * @param url The folder-relative url of the file.
@@ -50,9 +59,10 @@ export class _Files extends _SharePointQueryableCollection<IFileInfo[]> {
         const response = await spPost(Files(this, `add(overwrite=${shouldOverWrite},url='${escapeQueryStrValue(url)}')`), {
             body: content,
         });
+
         return {
             data: response,
-            file: this.getByName(url),
+            file: this.getUsingAbsoluteUrl(response["odata.id"]),
         };
     }
 
--- a/packages/sp/files/types.ts
+++ b/packages/sp/files/types.ts
@@ -8,7 +8,7 @@ import {
     deleteableWithETag,
 } from "../sharepointqueryable.js";
 import { TextParser, BlobParser, JSONParser, BufferParser, headers, body } from "@pnp/odata";
-import { assign, getGUID, isFunc, stringIsNullOrEmpty, isUrlAbsolute } from "@pnp/common";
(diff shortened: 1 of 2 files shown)


packages/sp/files/types.ts
```diff
@@ -37,6 +37,15 @@ export class _Files extends _SharePointQueryableCollection<IFileInfo[]> {
         return tag.configure(File(this).concat(`('${escapeQueryStrValue(name)}')`), "fis.getByName");
     }
 
+    /**
+     * Gets a File by file absolute url
+     *
+     * @param fileUrl The absolute url of the file
+     */
+     public getUsingAbsoluteUrl(fileUrl: string): IFile {
+        return tag.configure(File(`${fileUrl}`), "fis.getFileUsingResponseUrl");
```
wmertens, line 46: Should this be escaped and put into an alias to prevent path length issues?

aaademosu: I actually haven't encountered that issue before so didn't consider that. Reading up on it now and it seems to make sense. I would update the PR.

patrick-rodgers:
Hi @aaademosu - thank you for these two PRs! I had a look through the issue this is designed to address and have merged a different fix that should address the root cause. I want to leave these open in case my fix doesn't work for folks as we can circle back and review this proposal. Thank you again for taking the time to submit a fix / debug the library, very much appreciated!

As well we are getting closer to [v3 release](https://github.com/pnp/pnpjs/tree/version-3) and if you're interested in helping out more would love to get you going in that branch! We are a bit behind on dev docs, but encourage you to check out the code and share your thoughts.

aaademosu:
Hi @patrick-rodgers.

Would love to help. I would check out that branch and share my thoughts.

patrick-rodgers:
Going to close this as it appears the other fix was successful and as well is addressed in v3. Appreciate you taking the time and look forward to working with you more on v3 and beyond!

PR #2103, opened 2022-02-21
sp search v3 request body and getPage fix

SuperioOne:
#### Category
- [x] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?

**Search 400 bad request**
SPO search endpoint  doesn't accept search request in both browser(SPFx) and node environments.

> 400 Bad Request
> The parameter query does not exist in method postquery.

I updated request body similar to version-2 with `Content-Type : application/json;odata=verbose` header.

Search endpoint does not accept `application/json` and `application/json;odata=nometadata` as Content-Type header. It's probably expecting a different request body structure.

**SearchResults getPage has no observables to emit**
getPage function search call has no registered observables. I passed the copy of previous search instance so SearchResults can execute with correct timeline configuration.

### DIFF
--- a/packages/sp/search/query.ts
+++ b/packages/sp/search/query.ts
@@ -1,6 +1,6 @@
 import { _SPInstance, ISPQueryable } from "../spqueryable.js";
 import { hOP, isArray } from "@pnp/core";
-import { body } from "@pnp/queryable";
+import { body, InjectHeaders } from "@pnp/queryable";
 import { ISearchQuery, ISearchResponse, ISearchResult, ISearchBuilder, SearchQueryInit } from "./types.js";
 import { spPost } from "../operations.js";
 import { defaultPath } from "../decorators.js";
@@ -79,8 +79,6 @@ export function SearchQueryBuilder(queryText = "", _query = {}): ISearchBuilder 
     });
 }
 
-const queryRegex = /_api\/search\/postquery$/i;
-
 /**
  * Describes the search API
  *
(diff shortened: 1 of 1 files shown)


patrick-rodgers:
Hi @SuperioOne,

Thanks for this - a few quick questions/notes:

- The initial error "The parameter query does not exist in method postquery." seems to be from us not spreading the query (a bug) which you fixed. Thanks!
- I am not sure search only accepts verbose, I think the first error might when fixed might have then worked with minimal, but need to test on my end.
- I agree with you that the getPaged is missing observers - but need to test your fix/review it - this is another bug

If you remove the headers code in your PR does it still work? I'll give this a test when I have some time. Thanks again!!

SuperioOne:
Hi @patrick-rodgers ,

PR normally doesn't work when header is removed but I just found out the expected request format for `application/json`. Search Api expects array values instead of { results: prop }. The header can be removed, if fixArrProp function changed to something like this:

```typescript
    // packages/sp/search/query.ts:120
    private fixArrProp(prop: any) {
        if (typeof prop === "undefined") {
            return [];
        }

        return isArray(prop) ? prop : [prop];
    }
```

djsladi:
Any news when this fix is going to be available?

patrick-rodgers:
Thanks @SuperioOne, got this updated and tested this morning. Seach should be all set now, sorry again for the miss here, that's on us. Will get a release out shortly so this update is public.

PR #2144, opened 2022-03-07
Fix getAllChildrenAsOrderedTree retrieve properties

lch-dbrown:
#### Category
- [x] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?

#### Related Issues

Fixes an incomplete merge of V2 features into V3.

#### What's in this Pull Request?

This pull requests fixes getAllChildrenAsOrderedTree() method. This method now correctly retrieves term properties as specified by the parameter retrieveProperties.

PR #2158, opened 2022-03-14
Further address getAllChildrenAsOrderedTree.

lch-dbrown:
#### Category
- [x] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?

#### Related Issues
Additional to PR #2144. 

Another missed line from the getAllChildrenAsOrderedTree method. Admittedly, I missed the other select method within this function in the previous PR #2144. Also add tests to cover this method, with and without the additional prop _retreiveProperties_ .
- **getAllChildrenAsOrderedTree**: Tests that the method runs and returns an array.
- **getAllChildrenAsOrderedTree-retreiveProperties**: Tests that, if the method does return a term, the term has the localProperties key. This test would have caught the bug addressed by this PR and  #2144.


juliemturner:
Wonderful! Thanks so much for the added tests!

PR #2286, opened 2022-05-30
sp/views/types: add missing query escaping

wmertens:
#### Category

- [x] Bug fix?

Escape URL-embedded values in this particular file. Perhaps the better option would be to use aliases as well so that URL lengths are never an issue.

No effort has been done to check other files for this same issue.

This change is semi-breaking: If anybody was doing manual escapes, this will double-escape and break the code; others that didn't escape and didn't notice issues weren't embedding problematic characters and therefore won't notice issues.

IMHO, when one abstracts away the URL-based nature of the SP API, this also means abstracting away the details of its encoding, and making sure it works correctly for all inputs.

### DIFF
--- a/packages/sp/views/types.ts
+++ b/packages/sp/views/types.ts
@@ -8,6 +8,7 @@ import {
 } from "../spqueryable.js";
 import { defaultPath } from "../decorators.js";
 import { spPost, spPostMerge } from "../operations.js";
+import { escapeQueryStrValue } from "../utils/escape-query-str.js";
 
 @defaultPath("views")
 export class _Views extends _SPCollection<IViewInfo[]> {
@@ -48,7 +49,7 @@ export class _Views extends _SPCollection<IViewInfo[]> {
      * @param title The case-sensitive title of the view
      */
     public getByTitle(title: string): IView {
-        return View(this, `getByTitle('${title}')`);
+        return View(this, `getByTitle('${escapeQueryStrValue(title)}')`);
     }
 }
 export interface IViews extends _Views { }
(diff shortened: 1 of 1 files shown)


patrick-rodgers:
We're going to hold this PR for now. Because it is a breaking change and doesn't address an open or historical issue we want to think about it a bit more. Thanks!

wmertens:
I mean, I had to work around it, so if you want I can open an issue for it and point to this pr as the solution :)

juliemturner:
@wmertens - As @patrick-rodgers points out, this pr is the first we're hearing about an issue and therefore we want to take time to make sure the right decisions are made, especially, considering as you identify this is an incomplete and breaking change. I feel it's very important to think carefully about implementing anything that can be a breaking change as it's not a small undertaking as normally when you do so you would release a major version. Given all of these things, we believe this update requires more consideration before we take action.  

Although we very much appreciate your input and contribution, I would ask that you consider your tone. As unpaid maintainers of this library we do the absolute best we can while trying to juggle our work and personal lives. Please keep that in mind in your future comments.

wmertens:
Apologies, my tone was meant to be jocular not annoyed. I understand your position.

But do you agree with the premise of this PR, that the project should abstract away the details of the SharePoint API?

juliemturner:
On the surface neither do we which is why we didn't just close it outright.... but a breaking change is nothing to be thrown around without reflection.

patrick-rodgers:
Ok, we've revamped how we do querystring encoding to use the URLSearchParams class. As such we've update how we encode these path parts using a new function. I've included these updates in 3.6 so going to close this PR. Thanks!

PR #2287, opened 2022-05-30
sp/fields/types: add() properties can be undefined

wmertens:
#### Category
- [x] Bug fix?

properties of the fields.add() call can be undefined (as evidenced a few lines further down) but are being dereferenced to read `Required`.


juliemturner:
I don't think this modification is needed as `properties` is a required value of the method so as written should be correct. Do you have a specific issue that this change addresses?

wmertens:
As can be seen on line 77 of the file, the function expects `properties` to be possibly `undefined`.

I'm encountering a failure when calling `list.defaultView.fields.add('some title')`, which doesn't take any parameters.

patrick-rodgers:
Sure, but you should be getting a typing error as the second parameter is required by the method signature. `fieldTypeKind` is also required - which you aren't passing there. So yes, you can get errors if you call the methods incorrectly, but we are relying on the typings to handle most of those cases. Otherwise, we'd have to add tons of parameter checks into the library, and for all properties of complex parameters - which is wasted code space when the typings handle it.

juliemturner:
To be honest I'd rather see a fix to change addBoolean to make sure it provided a value... I really don't agree that we should be making those parameters optional.

wmertens:
In .add, the properties are treated as optional because when they are defined they have to do another call, extra work.

In .addField, the properties are optional by the nature of the `...properties` spread, so there again defining empty properties and then spreading them just causes extra work with no benefit.

PR #2291, opened 2022-05-31
sp/files copyByPath: fix falsy shouldOverWrite

wmertens:
#### Category
- [x] Bug fix?

fix the case where shouldOverWrite is undefined


juliemturner:
I'm confused about the need for this is because `shouldOverWrite` is a required value of the method and typed as a boolean. I would therefore think that this modification wouldn't be needed.

wmertens:
I encountered this when migrating from v1 to v3 and bulk-replacing `.moveTo` with `.moveByPath`, and the boolean used to be optional.

IMHO it's nicer if an API is more forgiving.

PR #2443, opened 2022-10-17
docs(graph--lists): add notice on how to get items

michael-ra:
#### Category
- [ ] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [x] Documentation update?

#### Related Issues

N/A

As there is no documentation about retrieving items from the lists.getByID object "IList" returned from the query:
- add a notice that links to other document file stating that you can query list items with call by path (this is only stated under "sites", which makes no sense, because anyone looking for items will look under lists as it's a property of "IList")

### DIFF
--- a/docs/graph/items.md
+++ b/docs/graph/items.md
@@ -1,0 +1,28 @@
+# @pnp/graph/items
+
+Currently, there is no module in graph to access all items directly. Please, instead, default to search by path using the following methods.
+
+[![Selective Imports Banner](https://img.shields.io/badge/Selective%20Imports-informational.svg)](../concepts/selective-imports.md)  
+
+### Get list items
+
+```TypeScript
+import { Site } from "@pnp/graph/sites";
+
+const sites = graph.sites.getById("{site id}");
+
+const items = await Site(sites, "lists/{listid}/items")();
+```
+
+### Get list items with fields included
+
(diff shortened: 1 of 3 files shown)


michael-ra:
#### Category of pull request additions
- [ ] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [x] Documentation update?

#### Related Issues

N/A

Refactored documentation, including dedicated items documentation, that replaces the code snippets at "call by path" in "@pnpjs/docs/graph/sites". Thus the documentation is more readable and you can find a way to retrieve items more easily with Graph API as a beginner.
Also linked "@pnpjs/docs/graph/lists" with the new items doc file, so that users looking there to retrieve items will know what to do.

juliemturner:
@michael-ra -- Thanks for pointing this out. Yes, our coverage of the graph is certainly lack luster and there is a history for why, but all that reasoning is moot at this point. Long story short, we have just started on the path to rework the implementation and coverage for the graph portion of the library, and I would have really loved to see someone take on the implementation of list items vs just shuffling around documentation in a way that's actually fairly inconsistent with our current method, but I can see the short-term benefit. So, I'll review this and get it merged in but hope that in the very near future we can replace it with the full implementation of List Items.  Thanks for the time and effort.

PR #2523, opened 2023-01-05
Version 2 package updates because of CVEs

dvTB:
#### Category
- [x ] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?

#### Related Issues
Fixed with node-fetch >=2.6.7
https://github.com/advisories/GHSA-8cf7-32gw-wr33 

fixed with jsonwebtoken >=9.0.0.0
https://github.com/advisories/GHSA-r683-j2x4-v87g  
https://github.com/advisories/GHSA-27h2-hvpr-p74q 
https://github.com/advisories/GHSA-hjrf-2m68-5959  
https://github.com/advisories/GHSA-qwph-4952-7xr6  

Just Package updates so the fixes can be applied.
node-fetch: "^2.6.7"
jsonwebtoken: "^9.0.0"
jsonwebtoken was updated to 9.0.0.0 the only fixed version but now incompatible with  Node 11 (support for that version is long gone)


juliemturner:
Hi @dvTB,

Thank you for your submission. As we've mentioned we're no longer supporting V2, but have been taking PRs and merging and making new releases as needed. That said, there is some breaking package in the chain that is not allowing for an update to the root packages package-lock.json. The changes you're proposing in this PR could be deployed because you did not alter that root package-lock.json however that would then create a scenario where the referenced versions of the packages you changed would not match and our feeling is that could be problematic.

PR #2533, opened 2023-01-13
setStreamContentChunked: await loading chunks

### LINKED ISSUE #2532: Uploading a file stream loads the entire file stream into memory
### Category
- [x] Bug

### Version

Please specify what version of the library you are using: [latest]

Please specify what version(s) of SharePoint you are targeting: [online]

*If you are not using the latest release, please update and see if the issue is resolved before submitting an issue.*

When using chunked uploading with a ReadStream, the file should be sent to sharepoint without issues.

### Observed Behavior

The entire file is read into memory while it's being uploaded

The reason is that the stream isn't paused while sending data

Cc @dzieni

Dzieni:
fixes #2532

#### Category
- [x] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?

#### Related Issues

fixes #2532

The previous version of the code was creating a promise chain in-advance for every chunk that has to be uploaded. Since every chunk has to be kept in memory until the promise is finished, it led to a situation in which the whole file content was kept in memory.

The solution is to pause the stream on every chunk and resume it after a successful push, so that the next chunk can be processed.

#### TODO

Tests

### DIFF
--- a/packages/nodejs/sp-extensions/stream.ts
+++ b/packages/nodejs/sp-extensions/stream.ts
@@ -43,20 +43,26 @@ extendFactory(File, {
 
             stream.on("data", (chunk) => {
 
+                stream.pause();
+
                 blockNumber += 1;
 
                 if (blockNumber === 0) {
 
-                    promise = promise.then(() => {
+                    promise = promise.then(async () => {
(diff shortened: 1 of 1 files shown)


patrick-rodgers:
Hey thanks for this, appreciate you submitting it - we're going to go ahead and merge it so it makes it into the 3.11 release going out today. Nice work!

PR #2540, opened 2023-01-17
Timeline: fix unhandlable promise in start()

### LINKED ISSUE #2539: Error handling: unhandled promise rejection despite proper catching
### Category
- [ ] Enhancement
- [x] Bug
- [ ] Question
- [ ] Documentation gap/issue

### Version

Please specify what version of the library you are using: 3.11

Please specify what version(s) of SharePoint you are targeting: all

```
const x = await sp.web.lists
	.getByTitle("This list surely does not exist")()
	.catch(() => "caught");
```

`x` should equal `"caught"`, no exceptions or warnings thrown by the runtime.

### Observed Behavior

`x` does equal `"caught"`, but unhandled promise warning is thrown.

Dzieni:
#### Category
- [X] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?

#### Related Issues

fixes #2539

Inside `start` method the `.execute` promise is wrapped in `.finally`, but this wrap was not returned. Therefore, the event loop had to handle 2 promises (wrapped and not-wrapped), and the wrapped one couldn't be handled properly in case of errors.

This PR fixes the problem by adding an empty `catch` to the wrapped promise.

### DIFF
--- a/packages/core/timeline.ts
+++ b/packages/core/timeline.ts
@@ -309,7 +309,7 @@ export abstract class Timeline<T extends Moments> {
 
                 this.error(e2);
             }
-        });
+        }).catch(() => void(0));
 
         // give the promise back to the caller
         return p;


wmertens:
To clarify, if the .finally() clause throws, it will propagate its error and supercede the original promise's rejection (if any).

So this will be an unhandled rejection if the p.finally isn't returned instead of p.

Dzieni:
@patrick-rodgers would it be possible to release that fix as 3.11.1?

juliemturner:
Given the 3.12 release is scheduled for Feb 10th I can't see the need for the added work to release a .1. We have [nightly builds](https://pnp.github.io/pnpjs/concepts/nightly-builds/) that you can use in the meantime.

PR #2546, opened 2023-01-23
docs: fix ConsoleListener creation

viceice:
#### Category
- [ ] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [x] Documentation update?

#### Related Issues

none

The docs still mention to use `new ConsoleListener()`, which is wrong and gives `Only a void function can be called with the 'new' keyword.ts(2350)` error in TypeScript.

https://github.com/pnp/pnpjs/blob/e757e59fb568aafcf5788558e3ffd1b212c9de1a/packages/logging/listeners.ts#L3-L5

So because that is now a wrapper function, it should be called as function.

### DIFF
--- a/docs/logging/index.md
+++ b/docs/logging/index.md
@@ -113,7 +113,7 @@ import {
 } from "@pnp/logging";
 
 // subscribe a listener
-Logger.subscribe(new ConsoleListener());
+Logger.subscribe(ConsoleListener());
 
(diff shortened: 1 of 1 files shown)


patrick-rodgers:
Nice catch, we obviously missed this update, thank you for noticing and submitting an update!!

PR #2551, opened 2023-01-25
Add typings when fetching Site data and change typings for getDocumentLibraries

jhholm:
#### Category
- [x] Bug fix?
- [x] New feature?
- [ ] New sample?
- [ ] Documentation update?

#### Related Issues

No related issues

When fetching site data with e.g. **const data = await sp.site()**, no typings existed in the return object. I added typings in a similar way as in the webs module. I also checked that the properties seem to be the same in a few standard release tenants.

Adding the typing required two **<any>** generics on queries on methods **getDocumentLibraries** and **getWebUrlFromPageUrl**.

I also noticed that **getDocumentLibraries** returning **IDocumentLibraryInformation** interface was missing some important properties, such as Id. I also changed the properties to be not optional, as the interface is only used as a return type.

### DIFF
--- a/packages/sp/sites/types.ts
+++ b/packages/sp/sites/types.ts
@@ -42,7 +42,7 @@ function rebaseSiteUrl(candidate: string, path: string | undefined): string {
 }
 
 @defaultPath("_api/site")
-export class _Site extends _SPInstance {
+export class _Site extends _SPInstance<ISiteInfo> {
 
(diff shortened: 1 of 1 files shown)


patrick-rodgers:
Thanks for making this contribution! Will be in tomorrow's nightly build as well in the next release 3.12. Thanks!

PR #2631, opened 2023-04-04
New Feature, Bug Fix, Documentation Update

### LINKED ISSUE #2629: Bug: Bad request when targeting specific item in recycle bin
### Category
- [ ] Enhancement
- [x] Bug
- [ ] Question
- [ ] Documentation gap/issue

### Version

Please specify what version of the library you are using: `v3.13.0`

Please specify what version(s) of SharePoint you are targeting: `SharePoint Online`

Trying to remove/restore an item in a SharePoint site recycle bin. It should restore/remove the item.

### Observed Behavior
Bad request:

image

This is because the ID should be surrounded in quotes: `recycleBin('b38e7b83-e405-4726-898f-774eda28543a')/Restore`

### Steps to Reproduce
```typescript
await sp.web.recycleBin.getById('<list item GUID>').restore();
```

juliemturner:
#### Category
- [x] Bug fix?
- [x] New feature?
- [ ] New sample?
- [x] Documentation update?

#### Related Issues

fixes #2629

- Bug fix for missing quotes on sp recycle bin getById method.
- Added new azidjsclient package that supports Azure Identity, most useful for node-js development and managed identity
- Updated debug scenarios for spa to make debugging work with settings file.
- Updated package-lock.json for SPFx debugging scenario to fix a bug
- Updated Taxonomy documentation to call out need for auth token on add/update/delete events.
- Update to node package GraphDefault and SPDefault package to make msal parameter optional (deprecated) and updated the documentation for new guidance on passing the MSAL (and AzureIdentity) auth methods as seperate observerable on the using call.

### DIFF
--- a/docs/concepts/auth-nodejs.md
+++ b/docs/concepts/auth-nodejs.md
@@ -1,0 +1,93 @@
+# Authentication in NodeJS
+
+We support MSAL for both browser and nodejs and Azure Identity for nodejs by providing a thin wrapper around the official libraries. We won't document the fully possible configurations, but any parameters supplied are passed through to the underlying implementation.
+
+Depending on which package you want to use you will need to install an additional package from the library because of the large dependencies.
+
+For the browser MSAL package:
+
+`npm install @pnp/msaljsclient --save`
+
+For the Azure Identity package:
+
+`npm install @pnp/azidjsclient --save`
+
+At this time we're using version 1.x of the `msal` library which uses Implicit Flow. For more informaiton on the msal library please see the [AzureAD/microsoft-authentication-library-for-js](https://github.com/AzureAD/microsoft-authentication-library-for-js#readme).
+
+The following samples reference a MSAL configuration that utilizes an Azure AD App Registration, these are samples that show the typings for those objects:
+
+```TypeScript
+import { Configuration, AuthenticationParameters } from "msal";
+
+const configuration: Configuration = {
+  auth: {
+    authority: "https://login.microsoftonline.com/{tenant Id}/",
+    clientId: "{AAD Application Id/Client Id}"
+  }
+};
+
+const authParams: AuthenticationParameters = {
+  scopes: ["https://graph.microsoft.com/.default"] 
+};
+```
+
+> Be sure to [review getting authentication setup](https://github.com/pnp/pnpjs/blob/version-3/docs/getting-started.md#authentication) for Nodejs.
+
+For nodejs we support MSAL through the [msal-node](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/README.md) library. Again we provide a thin wrapper to allow it to work within our request flow.
+
+## MSAL + NodeJS
+
+The SPDefault and GraphDefault exported by the nodejs library include MSAL and takes the parameters directly. Please consider that ability deprecated and instead use the method shown below to chain the MSAL auth behavior and configure it independently.
+
+```TypeScript
+import { SPDefault, GraphDefault } from "@pnp/nodejs";
+import { spfi } from "@pnp/sp";
+import { graphfi } from "@pnp/graph";
+import "@pnp/graph/users";
+import "@pnp/sp/webs";
+
+const msal = { config, scopes };
+
+const sp = spfi("https://{tenant}.sharepoint.com/sites/dev").using(
+    SPDefault(),
+    MSAL(configuration, authParams)
+);
+
+const graph = graphfi().using(
+    GraphDefault(),
+    MSAL(configuration, authParams)
+);
+
+const webData = await sp.web();
+const meData = await graph.me();
+```
+
+## Azure Identity + NodeJS
+
+This library provides a thin wrapper around the [@azure/identity](https://github.com/Azure/azure-sdk-for-js/blob/main/documentation/using-azure-identity.md) library which simplifies the authentication process and makes it easy to integrate Azure Identity authentication in your solution.
+
+```TypeScript
+import { DefaultAzureCredential } from "@azure/identity";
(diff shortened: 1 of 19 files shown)


docs/concepts/auth-spfx.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+## MSAL + SPFx
+
+```TypeScript
+import { spfi, SPBrowser } from "@pnp/sp";
+import { MSAL } from "@pnp/msaljsclient";
+import "@pnp/sp/webs";
+
+const sp = spfi("https://tenant.sharepoint.com/sites/dev").using(SPBrowser(), MSAL(configuration, authParams));
```
bcameron1231, line 48: me being picky - with the breakout of the auth pages, we're now missing the configuration and authParams initiation like we have on the other msal samples.

docs/concepts/auth-nodejs.md
```diff
@@ -0,0 +1,93 @@
...
+The SPDefault and GraphDefault exported by the nodejs library include MSAL and takes the parameters directly. Please consider that ability deprecated and instead use the method shown below to chain the MSAL auth behavior and configure it independently.
+
+```TypeScript
+import { SPDefault, GraphDefault } from "@pnp/nodejs";
+import { spfi } from "@pnp/sp";
+import { graphfi } from "@pnp/graph";
+import "@pnp/graph/users";
+import "@pnp/sp/webs";
+
+const msal = { config, scopes };
```
bcameron1231, line 49: Was this accidental? I don't see it being used in the sample.

bcameron1231 (commented):
Added comments to some of the docs.

docs/concepts/auth-spfx.md
juliemturner: Updated

docs/concepts/auth-nodejs.md
juliemturner: Updated

packages/azidjsclient/index.ts
```diff
@@ -0,0 +1,68 @@
...
+    UsernamePasswordCredential,
+    VisualStudioCodeCredential,
+} from "@azure/identity";
+
+export interface IAzureIdentityToken {
+    token: string;
+    expires: Date;
+}
+
+export type ValidCredential = DefaultAzureCredential | ChainedTokenCredential | EnvironmentCredential |
```
patrick-rodgers, line 28: Do we want to limit the initial set of credentials we call valid as part of this type? I don't know it well enough, but feels like a subset would apply to the library?

packages/azidjsclient/index.ts
```diff
@@ -0,0 +1,68 @@
...
+ManagedIdentityCredential | ClientAssertionCredential | ClientCertificateCredential | ClientSecretCredential |
+AuthorizationCodeCredential | DeviceCodeCredential | InteractiveBrowserCredential | OnBehalfOfCredential |
+UsernamePasswordCredential | AzureCliCredential | AzurePowerShellCredential | VisualStudioCodeCredential;
+
+export function AzureIdentity(credential: ValidCredential,
+    scopes: string[] = ["https://graph.microsoft.com/.default"],
+    options?: GetTokenOptions): (instance: Queryable) => Queryable {
+
+    return (instance: Queryable) => {
+        const key = `AzureIdentityCredential${getHashCode(scopes.join())}`;
```
patrick-rodgers, line 38: I'd move key and storage up into the parent function.

packages/azidjsclient/index.ts
```diff
@@ -0,0 +1,68 @@
...
+            if (tokenStore) {
+                const storedToken: IAzureIdentityToken = JSON.parse(tokenStore);
+                if (storedToken.expires < (new Date())) {
+                    token = storedToken.token;
+                }
+            }
+            if (token == null) {
+                const aiToken: AccessToken = await credential.getToken(scopes, options);
+                // Set expiration date equal to the expiration timestamp minus 5 minutes for buffer
+                expires = new Date((new Date()).getMilliseconds() + (aiToken.expiresOnTimestamp - 300000));
```
patrick-rodgers, line 55: do we need to create two dates here? dateAdd exists in core, can that help?

packages/azidjsclient/index.ts
```diff
@@ -0,0 +1,68 @@
...
+                token = aiToken.token;
+                const newToken: IAzureIdentityToken = { token, expires };
+                storage.session.put(key, JSON.stringify(newToken));
+            }
+
+            init.headers = { ...init.headers, Authorization: `Bearer ${token}` };
```
patrick-rodgers, line 61: Nothing to do with your code, but perhaps we need an addHeader func we can reuse as we've got this code in multiple places.

packages/azidjsclient/index.ts
juliemturner: I saw dateAdd, but it doesn't support milliseconds which is the format of the expiresOnTimestamp so I just did it by hand...

packages/azidjsclient/index.ts
juliemturner: Well... we could, but I'm not sure why we would given that this is just a wrapper... If someone has a use case that would work then removing it just creates a false barrier.

packages/azidjsclient/index.ts
bcameron1231: One potential option to not create a new date object, could possibly do `Date.now() % 1000` instead of `(new Date()).getMilliseconds()`

juliemturner: Wow guys, splitten some hairs here... but sure whatever.

patrick-rodgers:
Ship IT!

juliemturner:
Shipped!

PR #2689, opened 2023-05-21
Fix for#2688

HiltonGiesenow:
#### Category
- [x ] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?

#### Related Issues

Fixed #2688 

#### What's in this Pull Request?

Changes `attachmentFiles.add` to also encode the filename to ensure files with apostrophes in the filename (') can be uploaded.

PR #2693, opened 2023-05-24
Fixing sp fx testing webpart

HiltonGiesenow:
#### Category
- [x] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?

Fixes the broken SPFx test web part, per @bcameron1231 fix here: https://github.com/pnp/pnpjs/pull/2689


bcameron1231:
Thank you for the PR (and all the help)! I'm not sure this PR is really needed. While this does seem to fix the issue in the SPFx solution, we still have the underlying issue of the circular references. The circular references are what are causing the issues we're seeing here. Which is currently in code review, and should be fixed in the next release.

HiltonGiesenow:
ok, just putting it out there in case it can help, we can toss this PR if you think it's not worthwhile? If I can find some bandwidth, I'd like to help out on this SPFx testing harness soon, if that's ok?

bcameron1231:
Yea, I'd say it's okay to toss this. As for SPFx, ya that would be helpful. As Patrick mentioned, we don't advertise it, because it's not 100% yet, and we didn't want to publicize it if it had existing issues. But we would definitely welcome contributions :)

PR #2796, opened 2023-10-02
changed register custom request client factory to add runtime as parameter

### LINKED ISSUE #2795: registerCustomRequestClientFactory runtime cannot be passed to custom client

2.x (No longer supported)

### Minor Version Number

2.14.0 (Common-JS-Version)

### Target environment

NodeJS

I'm using node js and sp with createIsolated setup.

### Question/Request
We're using a custom request client like described in https://pnp.github.io/pnpjs/v2/sp/custom-irequestclient/. The problem is, that the runtime gets not passed to the custom request client. The runtime is needed, to use our SPFetchClient in the sp.setup method (see coding). Because if I pass the runtime to the super constructor (of SPHttpClient) the this._impl variable gets set right. Otherwise a FetchClient is simply set in the this._impl variable and as a result the method this._impl.fetch (in the fetchRaw method) does not work in the CustomSPHttpClient (runs on error)

That's the reason why I've created a new register method within this pull request to register a request client factory with a runtime.

The problem is that the following thing does not work.

```
// registering
registerCustomRequestClientFactory((runtime) => new CustomSPHttpClient(runtime));

// sp.setup
conn.setup({
	sp: {
	        baseUrl: baseUrl,
		fetchClientFactory: spFetchClient,
		headers,
	}
});

// CustomSPHttpClient
class CustomSPHttpClient extends SPHttpClient {
	constructor(runtime) {
		super(runtime); // here the runtime is always empty
	}

	fetchRaw(url, options) {
                 this._impl.fetch...
                 
	}
}
```

nob-b3:
#### Category
- [X] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?

#### Related Issues
fixes #2795. Another solution for the same problem is in #2797 

Problem is described here: #2795. Added the runtime parameter to the registerCustomRequestClientFactory method. So now the runtime is also a parameter in the constructor of the custom request client factory.

### DIFF
--- a/packages/sp/operations.ts
+++ b/packages/sp/operations.ts
@@ -4,8 +4,8 @@ import { ISharePointQueryable } from "./sharepointqueryable.js";
 import { IFetchOptions, mergeOptions, objectDefinedNotNull, IRequestClient, isFunc, Runtime } from "@pnp/common";
 import { toAbsoluteUrl } from "./utils/toabsoluteurl.js";
 
-export function registerCustomRequestClientFactory(requestClientFactory: () => IRequestClient) {
-    httpClientFactory = isFunc(requestClientFactory) ? () => requestClientFactory : defaultFactory;
+export function registerCustomRequestClientFactory(requestClientFactory: (runtime: Runtime) => () => IRequestClient) {
+    httpClientFactory = isFunc(requestClientFactory) ? requestClientFactory : defaultFactory;
 }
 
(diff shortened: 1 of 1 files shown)


nob-b3:
If the fix isn't correct. Would it be possible to create another method like registerCustomRequestClientFactoryWithRuntime to add the functionality?

PR #2797, opened 2023-10-02
added new register custom request client factory with runtime as parameter

### LINKED ISSUE #2795: registerCustomRequestClientFactory runtime cannot be passed to custom client

nob-b3:
#### Category
- [X] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?

#### Related Issues
fixes #2795. I've already created another pull request #2796 for this problem. This is just another possible solution for that.

Problem is described here #2795. The pull request creates a new method registerCustomRequestClientFactoryWithRuntime which enables to set a custom request client factory with the runtime as parameter.

### DIFF
--- a/packages/sp/operations.ts
+++ b/packages/sp/operations.ts
@@ -8,6 +8,10 @@ export function registerCustomRequestClientFactory(requestClientFactory: () => I
     httpClientFactory = isFunc(requestClientFactory) ? () => requestClientFactory : defaultFactory;
 }
 
+export function registerCustomRequestClientFactoryWithRuntime(requestClientFactory: (runtime: Runtime) => () => IRequestClient) {
+    httpClientFactory = isFunc(requestClientFactory) ? requestClientFactory : defaultFactory;
+}
+
 const defaultFactory = (runtime: Runtime) => () => new SPHttpClient(runtime);
 let httpClientFactory: (runtime: Runtime) => () => IRequestClient = defaultFactory;
 


nob-b3:
@bcameron1231 @juliemturner @patrick-rodgers 
Could you take a look at that because it's an important feature for us. Thank you in advance.

Kind regards, Noah

PR #2831, opened 2023-11-17
Use @azure/msal-browser instead of MSAL

datyin:
#### Category
- [ ] Bug fix?
- [x] New feature?
- [ ] New sample?
- [x] Documentation update?

#### Related Issues

Since MSAL is deprecated now, we should use @azure/msal-browser

Refactored MSAL behavior


juliemturner:
Yes, we're aware of this change but we would not want to make it in v3, we'll be updating it in v4 release. If you would like to resubmit it targeting the v4 branch that would be greatly appreciated.

PR #2857, opened 2023-12-12
Fix: fromAbsolutePath methods reference undefined `this`

RoelVB:
#### Category
- [x] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?

#### Related Issues

No related issue.

When creating a File/Folder object based on the absolute path you would get the error:
`Cannot read property '_url' of undefined`, because the `init` value here would be `undefined`:
https://github.com/pnp/pnpjs/blob/e62c30d92b8e6cae63baac8cd3fcd3f391c2dcc2/packages/queryable/queryable.ts#L80-L81

The methods `fileFromAbsolutePath`/`folderFromAbsolutePath` now use the `base` argument

### DIFF
--- a/packages/sp/files/types.ts
+++ b/packages/sp/files/types.ts
@@ -543,7 +543,7 @@ export function fileFromServerRelativePath(base: ISPQueryable, serverRelativePat
  */
 export async function fileFromAbsolutePath(base: ISPQueryable, absoluteFilePath: string): Promise<IFile> {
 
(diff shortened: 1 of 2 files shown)


patrick-rodgers:
Thanks for submitting this fix! Can't believe no one caught it before, appreciate you taking the time to create the PR. Fix will go out in tonight's nightly build as well as the release planned for Jan 15th.

Thank you!

Issue #3016, opened 2024-04-25
Adding Excel workbook support


ChapC:
### Target environment

All

### Enhancement Idea

Hi there!

I've been working with PnPJS to interact with Graph APIs for a few months now (thanks, btw!) and I recently found the need to use some of the Excel Workbook related endpoints. I couldn't see a library package for these operations, so I started implementing a few of the endpoints under a 'workbook' extension on DriveItem. 

At the moment I only have a few of the most simple operations (creating sessions, listing rows, tables), but I'll likely need to create more. I was wondering if this is something you guys would accept a PR for? The API doesn't seem overly complicated, so I'd be open to typing out the workbook, table and rows sections to start. Just thought I'd check to see if it's up for grabs, or if your team is already on top of it.

Thanks again for all your work on this library. Finding these kind of client libraries for an API is a dream :)

juliemturner:
Absolutely, we'd love to see a PR for this. That said we would not implement it directly in drive item. You would create a new folder for workbook and then extend drive item. You can see examples of this for other endpoints sortof like [lists](https://github.com/pnp/pnpjs/tree/version-4/packages/graph/lists) but since workbook only exists off drive items then you really could implement the pattern that's in `drive.ts` directly in `index.ts`... although it's fine to separate them too...

ChapC:
Great! I've created a draft pull request with some further questions and notes in there 👍

PR #3019, opened 2024-04-26
Workbook package

ChapC:
### Category

- [ ] Bug fix?
- [x] New feature?
- [ ] New sample?
- [ ] Documentation update?

### Related Issues

Mentioned in #3016 

Initial draft of Excel workbook support. 
This first commit includes
- workbook session creation
- listing/getting tables
- listing/getting rows

I'll implement more endpoints as I go, but have a couple of notes/questions I thought I'd post here first.

1. Your contributing guidelines mention I should skip writing tests for endpoints that don't support app auth. According to the docs, that's all of the workbook endpoints. Is there any way I can write tests we can use for these endpoints?
2. In my testing, I wasn't able to use workbook operations on items opened by path. These failed with "Could not obtain a WAC token". When I use getItemById, it works. I wrote this off as a Graph API bug, but haven't tested extensively.
3. My current implementation of workbook sessions involves a `workbook` property on DriveItem that leads to a regular, sessionless IWorkbook and an async method `getWorkbookSession` that resolves to an IWorkbookWithSession. The latter is an extension of IWorkbook that adds session refresh/close methods, and uses the InjectHeaders behaviour to add the session id header. Thoughts on this approach?
4. I've added a prefer-async behavior to the graph folder. This is to handle the long-running operation pattern that a few of the workbook endpoints support. Again, I'm not sure if this is the ideal approach, but it's what I've got at the moment.

Welcoming all feedback. Thanks!


packages/graph/behaviors/prefer-async.ts
```diff
@@ -0,0 +1,39 @@
...
+
+        instance.on.parse(async function (url, response, result) {
+            if (response.status === 202) {
+                const opLocation = response.headers.get("Location");
+                const opId = opLocation.split("/").at(-1);
+
+                let succeeded = false;
+                const statusQuery = GraphQueryable(instance, `operations/${opId}`);
+                for (let i = 0; i < maxPolls; i++) {
+                    await new Promise(resolve => setTimeout(resolve, pollIntervalMs));
```
patrick-rodgers, line 18: We have [delay()](https://github.com/pnp/pnpjs/blob/version-4/packages/core/util.ts#L180) in core you can use for this very thing.

packages/graph/behaviors/prefer-async.ts
```diff
@@ -0,0 +1,39 @@
...
+
+                    const status = await graphGet<RichLongRunningOperation>(statusQuery);
```
patrick-rodgers, line 20: For GET you should be able to invoke the queryable itself `await statusQuery<>()`

packages/graph/behaviors/prefer-async.ts
```diff
@@ -0,0 +1,39 @@
...
+import { InjectHeaders } from "@pnp/queryable";
+import { GraphQueryable, _GraphQueryable, graphGet } from "../graphqueryable.js";
+import { RichLongRunningOperation } from "@microsoft/microsoft-graph-types";
+
+export function PreferAsync(pollIntervalMs: number = 25000, maxPolls: number = 4): TimelinePipe {
+    return (instance: _GraphQueryable) => {
+        instance.using(InjectHeaders({ "Prefer": "respond-async" }));
+
+        instance.on.parse(async function (url, response, result) {
+            if (response.status === 202) {
```
patrick-rodgers, line 11: What happens if you get a different response? We have some standard logic [wrapped up here](https://github.com/pnp/pnpjs/blob/version-4/packages/queryable/behaviors/parsers.ts#L100) to check for errors. Used [here](https://github.com/pnp/pnpjs/blob/version-4/packages/queryable/behaviors/parsers.ts#L7) and [here](https://github.com/pnp/pnpjs/blob/version-4/packages/queryable/behaviors/parsers.ts#L23) for example.

packages/graph/workbooks/session.ts
```diff
@@ -0,0 +1,20 @@
+import { GraphQueryable, graphPost } from "../graphqueryable.js";
+import { InjectHeaders, body } from "@pnp/queryable";
+import { IWorkbookWithSession, WorkbookWithSession } from "./types.js";
+import { _DriveItem } from "../files/types.js";
+import {
+    WorkbookSessionInfo
+} from "@microsoft/microsoft-graph-types";
+
+export async function getWorkbookSession(this: _DriveItem, persistChanges: boolean): Promise<IWorkbookWithSession> {
```
patrick-rodgers, line 9: If the only place this will really get used is in the property in the other file you can code it there directly. i.e. remove the export here and put the code there. But if this is a function that will be used on its own this makes sense.

patrick-rodgers (commented):
Really nice work! You did a great job figuring out and integrating with the library. I left a few comments for you to consider. Excited for your contribution!!

patrick-rodgers:
**1. Your contributing guidelines mention I should skip writing tests for endpoints that don't support app auth. According to the docs, that's all of the workbook endpoints. Is there any way I can write tests we can use for these endpoints?**

Not really. You can test them locally. We are working on test recording to allow us to record delegated-auth tests and replay them to validate.

**2. In my testing, I wasn't able to use workbook operations on items opened by path. These failed with "Could not obtain a WAC token". When I use getItemById, it works. I wrote this off as a Graph API bug, but haven't tested extensively.**

Don't know, WAC is the office protocol, perhaps that endpoint requires something special.

**3. My current implementation of workbook sessions involves a `workbook` property on DriveItem that leads to a regular, sessionless IWorkbook and an async method `getWorkbookSession` that resolves to an IWorkbookWithSession. The latter is an extension of IWorkbook that adds session refresh/close methods, and uses the InjectHeaders behaviour to add the session id header. Thoughts on this approach?**

I am not deeply familiar with this API, what is the difference between sessionless and with session? Do we need to expose both?

**4. I've added a prefer-async behavior to the graph folder. This is to handle the long-running operation pattern that a few of the workbook endpoints support. Again, I'm not sure if this is the ideal approach, but it's what I've got at the moment.**

This is pretty slick, I left a few notes inline. This isn't something we built-in for v4 launch, but I can see a ton of value to including it.

Again, nice work!

bcameron1231:
HI @ChapC is this ready? We see your changes, we were just waiting on anymore feedback from you.

ChapC:
Hey @bcameron1231, I've just added a few more of the endpoints I've tested recently. I was hoping to cover much more of the workbook functionality, but I've ended up with much less time than I thought I would have on this. Apologies!

So far, of [all the workbook sections](https://learn.microsoft.com/en-us/graph/api/resources/excel), I've implemented the basic stuff in Tables, Worksheets, Rows/Columns and Ranges. Everything to get/update/delete data in the spreadsheet. I have mocha tests for most of these on my local machine, although they're in a separate project as I had trouble getting it working with this one.

There were a couple quirks I encountered in these endpoints. Most I mentioned previously, but the biggest one I couldn't find a fix for was incompatibility with paging. The async iterator doesn't work for table rows, because as far as I can tell the Graph API simply doesn't return `hasNext`, `count`, etc. for the `workbook/{table name|id}/rows` endpoint. I've replicated this in Graph explorer as well. You can add `top` and `skip`, just fine, but `count` has no effect and the responses never include `@odata.nextLink`.

There's still quite a bit more for full coverage of all endpoints, like charts, running formulas and cell styling. I'm happy to keep chipping away at these features, just won't have a timeframe for you unfortunately. I can get some documentation written up soon if you guys decide you'd like to merge just the stuff I've got working rather than the full set. Just let me know :)

ChapC:
Also, to answer the question from @patrick-rodgers regarding session vs sessionless; my understanding is that both work for pretty much every endpoint, but using a session leads to better performance. [It's briefly explained in the docs here](https://learn.microsoft.com/en-us/graph/excel-manage-sessions). To quote them directly:

> The session header is not required for an Excel API to work. However, we recommend that you use the session header to improve performance. 

It also allows you to create non-persistent sessions, in which you can edit the workbook and run calculations without your changes being saved to the file. In sessionless mode, your changes are always immediately saved to the file.

patrick-rodgers:
Sounds good @ChapC! No worries, we're all busy so appreciate any time you have to contribute 😀 Let us know once you're ready for a full review. Thank you!!

ChapC:
After a few more bursts of work, I've finished the majority of endpoints and now have something I think could be worth merging! The last group of endpoints still missing are the Chart-related ones (chart creation, data selection, styling). For now, I think I'll leave those ones as a neat package for another PR down the road (either from myself or someone else who's keen).

The other thing I still need to write is documentation. As you can see from the [MS docs](https://learn.microsoft.com/en-us/graph/api/resources/excel), there are quite a few sections to cover. I'm planning on having most of them just point straight to the docs and adding a few examples for common tasks/quirks.

There are also a few places where the official docs appear to be incorrect. I've noted these as comments in the code where applicable. The biggest of those quirks are the ones I mentioned earlier in this thread: async paging doesn't seem to work and you must open a workbook on an DriveItem ID rather than path. If it's something that's out of our hands, I'll just leave a note in the documentation.

Let me know what you think!

patrick-rodgers:
@ChapC  - this is awesome! On the docs - you don't need to go too crazy - really just pointers on how to use THIS library. No need to document the full feature, that's the product's job. You can check out our other docs for how we've approached it - very minimal just to get folks started.

One small nit we found doing another review is the `getItemAt` decorator seems to only be Excel specific? If so, let's move that into the types.ts for excel.

Once you have the docs this looks really solid, excited to get it merged!! Thank you for taking the time to contribute 🙂

juliemturner:
Hi @ChapC -- Looks like you addressed the modifications we requested, are we ready to merge this in?

ChapC:
Thanks for checking in on this Julie. Happy new year!

Feature-wise we're all set, I just haven't had time to write the docs yet.

I'm hoping to have a few hours to spend on this in the next couple of weeks. Shouldn't take too long as I'm mostly pulling example code snippets from my local tests.

juliemturner:
Sounds good, thanks for the update.

ChapC:
I think I'm all wrapped up on the docs ✅

juliemturner:
@ChapC  - Awesome, should we move it in then, it's still marked draft so I just want to make sure you're ready.

ChapC:
Oops, yes sorry! Marked as ready.

juliemturner (dismissed):
Updated package-lock.json to match v4 branch.

juliemturner:
@ChapC - Hi. I approved the update (after fixing the outdated package-lock.json -- in the future best to avoid checking in those foundational files) but you apparently have linting errors. I could fix them for you but that would mean that I would end up having to close this pr and merging it in as my own. Let me know if you'd like to update the issues the linter found. You can run your own lint check locally by running `npm run lint`

ChapC:
Thanks for the heads up @juliemturner. Just ran eslint fix mode and your lint check seems to be happy on my end ✨
Sorry I missed that step from your guidelines!

juliemturner (approved):

juliemturner:
No worries, we've all done it a million times too. Thanks for your contribution!

Issue #3096, opened 2024-07-24
MC791596 impact on save page


seblhe:
### Major Version

4.x

### Minor Version Number

1.0

### Target environment

All

### Additional environment details

I'm using NodeJs (18.20.3) and PnPjs(4.3.0)

### Expected or Desired Behavior

With the MC791596 the Title area is not mandatory since july, there is a new webPart Banner with the same wePart Id as Title area.
If I a create a page with the new webPart Banner which is not mandatory and then use PnPJs to save the page, 

### Observed Behavior

PnPJs add the Title area webpart so I get 2 page headers.
Which is surprising is that we can't get with PnPjs the Title Area webpart added with PnPJs

### Steps to Reproduce

import { spfi } from "@pnp/sp";
import { IClientsidePage } from "@pnp/sp/clientside-pages";

const sp = spfi(...);

// our page instance
const page: IClientsidePage = await sp.web.loadClientsidePage("/sites/dev/sitepages/home.aspx");

// get the current value
const value = page.bannerImageUrl;

// set the value
page.bannerImageUrl = "/server/relative/path/to/image.png";
await page.save();

juliemturner:
Thanks @seblhe that said we were waiting for an official pages API and although we got one it doesn't have all the features it needs to have so we're sadly still waiting. As we outline in our docs the solution we have is reverse engineered from the first party  APIs and will break if anything changes on Microsoft's side. As you see that's coming to fruition which is why reverse engineering anything is a bad idea. Sadly, maintaining this code take a lot of work, and the team is very busy currently. I'll discuss it with the team and we'll see if/when we can take a look at this issue. Obviously, we're happy to take PR's from the community.

Bobinho4711:
HI @seblhe, did you maybe find an workaround for this issue?

juliemturner:
This issue all revolves around first party APIs, specifically the Pages API that currently in PnPjs is a reverse engineered version of those 1st party APIs, for more details see this [documentation](https://pnp.github.io/pnpjs/concepts/adv-clientside-pages/). At this point since a less robust publicly supported pages API has been GA'd we are in a holding pattern. We could remove our implementation and implement the public one but that will mean removing a lot of the functionality. If someone wants to take a stab at it, PR's are welcome.

juliemturner:
SharePoint Client-side Pages API is being deprecated and no changes will be made to the existing endpoint. Closing this issue to track pages updates to #3137

PR #3142, opened 2024-09-27
Workaround for MC791596

### LINKED ISSUE #3095: PnPJs "setBannerImage" is not updating the bannerImageUrl for the sitepages

2.x (No longer supported)

### Minor Version Number

2.11.0

### Target environment

SharePoint Framework

I'm using nodejs (14.21.3) and PnPjs(2.11.0)

### Question/Request

Hi,

I've been trying to update the bannerImageUrl for a site page using PnPJs, but the bannerImage is not updating.

We need to copy an existing page on the site. While the copying process works, the banner image is missing on the copied page. To fix this, I attempted to update the banner image using the following code:

```
sp.web
      .loadClientsidePage("<url of the page>")
      .then((page) => {
page
          .copy(sp.web, _title, _title, false)
          .then((res: any) => {
 res.setBannerImage(
              "/sites/siteName/SiteAssets/sitepages/Afternoon-News/image.jpg",
              {
                altText: "Image description",
                imageSourceType: 2,
                translateX: 30,
                translateY: 1234,
              }
            );
res.save();
window.open(`${res.json.AbsoluteUrl}?Mode=Edit`, "_blank");
})
})
```
However, it's not working as expected. Could you help me resolve this issue?
I am using PnPJs version 2.11.0. I've encountered the same issue reported here: the bannerImageUrl for a site page is not updating.

jhholm:
### Category

- [x] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?

### Related Issues

Fixes #3095 + #3096  and related

I changed the logic on updating the banner data without introducing breaking changes. It's not a 100% proper fix, but should work for most use cases.

### Further information

I reverse engineered how the banner rework works. The old way of using a banner keeps the data in LayoutWebpartsContent field. The banner rework seems to keep that field as an empty array. Pages with the old banner format are migrated in browser when editing the page.

I created a simple wrapper which updates either the old way of using a banner or the new way of using a banner as a separate webpart.

I noticed that when modifying a client side page in browser with multiple banners, only the first banner is being updated so I mimicked that in the wrapper logic.

Tested the following use cases

1. Create a clientside page with pnpjs and updating the page with pnpjs
    - This will keep the page with the old banner in LayoutWebpartsContent
    - You can normally edit the page in browser, which whill migrate to the new banner webpart
2. Create a clientside page with pnpjs, update the page manually in browser
     - Editing in browser will update the page from the old banner format the new banner webpart
3. Create a clientside page in browser with a banner
     - Will only edit the new banner webpart, and not save anything to LayoutWebpartsContent
4. Modifying a page with multiple new banner webparts
     - Should only set the changes on the first banner on the page, as does updating the title of the page in a browser

These use cases will not work, and would probably require breaking changes
1. Create a clientside page in browser and remove the new banner, update page with pnpjs
    - Logic will add the old banner
2. Modifying a clientside page with pnpjs that has an old banner and a new banner
    - I don't see this as a viable real use case
    - Will actually edit the new banner and not the old banner

Tested with all the banner modifications (changing image, title, etc). I created this for version-3, as that's what I'm still using on my own projects.

### DIFF
--- a/packages/sp/clientside-pages/types.ts
+++ b/packages/sp/clientside-pages/types.ts
@@ -15,6 +15,11 @@ import "../comments/item.js";
 import { createBatch } from "../batching.js";
 
 /**
+ * Id of the Banner Web Part
+ */
+const BannerWebPartId = "cbe7b0a9-3504-44dd-a3a3-0e5cacd07788";
+
+/**
  * Page promotion state
  */
 export const enum PromotedState {
@@ -109,6 +114,28 @@ export class _ClientsidePage extends _SPQueryable {
         };
     }
 
+    /**
+     * Returns either the old this._layoutPart or the first banner webpart on the page
+     */
+    private get _bannerWrapper() {
+        const bannerWebPart = this.findControl(control => control.data.webPartId === BannerWebPartId);
+        if (bannerWebPart) {
+            return bannerWebPart.data as ILayoutPartsContent;
+        }
+        return this._layoutPart;
+    }
+
+    /**
+     * Returns either the old this._layoutPart.properties or the first banner webpart properties on the page
+     */
+    private get _bannerWrapperProperties() {
+        const bannerWebPart = this.findControl(control => control.data.webPartId === BannerWebPartId);
+        if (bannerWebPart) {
+            return bannerWebPart.data.webPartData.properties as ILayoutPartsContent["properties"];
+        }
+        return this._layoutPart.properties;
+    }
+
     public get pageLayout(): ClientsidePageLayoutType {
         return this.json.PageLayoutType;
     }
@@ -141,7 +168,7 @@ export class _ClientsidePage extends _SPQueryable {
 
     public set topicHeader(value: string) {
         this.json.TopicHeader = value;
-        this._layoutPart.properties.topicHeader = value;
+        this._bannerWrapperProperties.topicHeader = value;
         if (stringIsNullOrEmpty(value)) {
             this.showTopicHeader = false;
         }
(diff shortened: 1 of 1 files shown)


heesungjang:
Hi @juliemturner ,

This issue is impacting one of our core product features, and I'm trying to resolve this from my end since this API is now deprecated, at least until the new API in Graph is available. Could I ask for your advice on possible approaches to create a wrapper class for the workaround? I can't think of a way to apply this workaround pull request other than forking the library (not even sure if that's possible). 

I'm sure your days have been very busy, but when you have time, it would be really helpful to get your thoughts on this.

Thank you.

jhholm:
Please note that my proposed change is not a proper fix and there are use cases where it doesn't work. You are more than welcome to test and use my fork if you want, knowing that it might not work for you.

I would recommend looking into [patch-package](https://www.npmjs.com/package/patch-package) or if you're using pnpm [pnpm patch <pkg>](https://pnpm.io/cli/patch).

The workflow is about the following:

1. Preferably create your own fork of pnpjs, and apply the changes you need
2. npm run build (for your pnpjs fork)
3. npm run package (for your pnpjs fork)
4. Install the same exact version of pnpjs to your project that you are using
5. Copy the changes from your fork of pnpjs /dist to your node_modules @pnp/sp folder
6. npx patch-package @pnp/sp
7. Add the postinstall script "patch-package"

Note that there is only one file changed in this pull request. Also note that the file is different between versions 3.x and 4.x. Should be trivial to port to 4.x.

PR #3155, opened 2024-10-21
OData lambda style filter query

Tanddant:
### Category

- [ ] Bug fix?
- [x] New feature?
- [ ] New sample?
- [ ] Documentation update?

### Related Issues

Based off the discussion over on #2781

This PR is not completely done, but I needed a place to track the final changes (and docs) based off the input in #2781 

 - [x] We like "method 1", seems simpler
 - [x] Drop "Field" from all the names so Lookup() vs LookupField().
 - [x] switch to camel case so Lookup becomes lookup()
 - [x] could simplify the names, equal instead of EqualsTo and notEqual vs NotEqualTo
     - I'm still open to the idea of having even shorter values, such as `.eq(...)` alongside the others, would like opinions 
 - [x] Swap and/or for All/Some, just seems clearer to someone showing up with no knowledge of the filter methods
     - Not sure I'm the biggest fan, see #2781 for the reasons why
 - [x] pass in multiple objects you can use ...[] to gather them into an array within the method.
 - [x] this should be able to support all the filter cases, another reason to drop the "Field" name from the builder functions. Like for a site if you give it ISiteInfo you could filter on Text("Title").EqualsTo("blah");
     - The following is tested to work 
     ```TypeScript
     sp.web.lists.filter<IListInfo>(f => f.text("Title").equal("Employees"))
     ```

I'll happily start updating the docs once we have a "somewhat sign-off" on the above been done to an acceptable state 😊


Tanddant:
Latest commit added "auto typing" for things like lists, webs, and so on, where we "know the type"

I.e
```TypeScript
await sp.web.webinfos.filter(w => w.text("Title").equal('MyWebTitle'))();
```
Without having to specify that `Title` is a field on lists

I haven't tested this in every possible combination, but it looks to be working without breaking the list item stuff, should greatly improve the usability 😊

bcameron1231:
Thanks! Will be doing some testing on this.

Tanddant:
@bcameron1231 Let me know if you have any issues, I have some initial work on the first draft of the docs, do you want them here, or in a separate PR?

bcameron1231:
Let's get those in the same commit. It will also help me test correctly. 😁 Excited to see this. 

Can you also try and squash your commits?

Tanddant:
So this might just be my git skills that are lacking - I can't seem to squash the old commits without ending up with like 300 merge confilts that need manual processing, but I at least got the newest ones squashed into two

bcameron1231:
Thank you. I've done some testing, it looks to be working. However it looks likes the plurals may be gone? i.e, .equal() vs .equals(). We'd like the plurals back 👍 

Here is what we plan to do once you make those changes. Given this has backwards compatibility, we're going to get this merged in and try to push it out in November. Our goal is to promote this new functionality in the next Month's call as a _beta_ functionality. 

Our hope is that we can get others to start using it and gather some feedback on the implementation. Maybe we could get you set up to do a Demo on a call at some point to explain how it works.

Tanddant:
I think the rename was just down to misunderstanding one of the last feedback items 😊 - changed it up for `.equals` `.notEquals()` `.greaterThanOrEquals()` `.lessThanOrEquals()`

Looking forward to getting some real world feedback once it hits beta!

juliemturner (approved):

bcameron1231:
🚀🚀🚀🚀🚀🚀

Tanddant:
Are the lint rules different on the runs here and the settings in the project? 😅 - I only had one warning locally, which I couldn't get rid of 👀

PR #3188, opened 2025-01-10
Add `drive` endpoint to lists in Graph

autinerd:
### Category

- [ ] Bug fix?
- [X] New feature?
- [ ] New sample?
- [ ] Documentation update?

This adds the `drive` endpoint to Graph lists, so that the corresponding Drive can be fetched from a list.

### DIFF
--- a/docs/graph/lists.md
+++ b/docs/graph/lists.md
@@ -81,6 +81,18 @@ const graph = graphfi(...);
 await graph.sites.getById("{site identifier}").lists.getById("{list identifier}").columns();
 ```
 
+## Get the corresponding drive to a list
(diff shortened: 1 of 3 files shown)


juliemturner:
Hi @autinerd 

Thank you very much for your contribution. In general your PR was well done and we appreciate it. Unfortunately you were returning IDrive which wasn't the right choice for this endpoint because it's not a chainable endpoint so we needed to return the Graph type for Drive. I took your PR and adjusted it #3190 , and also added an additional test for the new drive endpoint. Again, thank you for your contribution, we very much appreciate it.

PR #3202, opened 2025-02-01
fix content-type order type on folder

ltdu:
### Category

- [x] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?

This PR updates types for ```Folder.ContentTypeOrder``` and ```Folder.UniqueContentTypeOrder``` properties from ```string``` array to array of ```StringValue``` objects.

At least at some time it was described correctly: https://pnp.github.io/pnpjs/v1/sp/docs/content-types/

### DIFF
--- a/packages/sp/folders/types.ts
+++ b/packages/sp/folders/types.ts
@@ -343,8 +343,8 @@ export interface IFolderInfo {
     TimeLastModified: string;
     UniqueId: string;
     WelcomePage: string;
-    ContentTypeOrder: string[];
-    UniqueContentTypeOrder: string[];
+    ContentTypeOrder: { StringValue: string }[];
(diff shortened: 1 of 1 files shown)


patrick-rodgers:
@ltdu - help me understand, you've linked to the v1 docs which I believe at the time we were using still verbose which would have the format you indicate. Are you still seeing that? What is the request you are making to see that response?

ltdu:
Even with minimal and no metadata whose properties return ```{ StringValue: string }[]```, not simple ```string[]```.

patrick-rodgers:
AH, yep, you're right. Nice catch, thanks!

patrick-rodgers (approved):

PR #3335, opened 2026-04-05
Add recycle-bin to sp presets all

### LINKED ISSUE #3333: Bug: Recycle Bin module is missing from @pnp/sp/presets/all
### Major Version

4.x

### Minor Version Number

4.18.0

### Target environment

All

Noticed this bug with SP Editor when using PnP JS Console. Technically SP Editor seems to build their @pnp/sp bundle with the code paths provided by @pnp/sp/presets/all. Thus even though I would explicitly import "@pnp/sp/recycle-bin" additionally, the code would not work. Technically this is a downstream bug in SP Editor, not PnP JS.

This isn't an issue in most custom projects when you are using selective imports. In any case you will need to always selectively import @pnp/sp/recycle-bin if you want to use that module.

### Expected or Desired Behavior

When using 
```
import "@pnp/sp/presets/all"
```
I'm expecting all modules to be loaded with this preset.

### Observed Behavior

Recycle Bin module is not loaded and is not available if not explicitly loaded with
```
import "@pnp/sp/recycle-bin"
```

### Steps to Reproduce

import { spfi } from "@pnp/sp";
import "@pnp/sp/presets/all";

const sp = spfi(...);

//You should get a typescript error here
const bin = await sp.web.recycleBin();

jhholm:
### Category

- [X] Bug fix?

### Related Issues

fixes #3333

Adds recycle-bin to sp presets all

### DIFF
--- a/packages/sp/presets/all.ts
+++ b/packages/sp/presets/all.ts
@@ -18,6 +18,7 @@ import "../lists/index.js";
 import "../navigation/index.js";
 import "../profiles/index.js";
 import "../publishing-sitepageservice/index.js";
+import "../recycle-bin/index.js";
 import "../regional-settings/index.js";
 import "../related-items/index.js";
 import "../search/index.js";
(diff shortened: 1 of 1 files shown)


juliemturner (approved):

patrick-rodgers:
Thanks @jhholm!

PR #3351, opened 2026-06-05
Version 5: Updates for messages and chat plus tsconfig fixes.

juliemturner:
### Category

- [x] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?

Updates for v5


packages/graph/users/types.ts
```diff
@@ -47,7 +48,11 @@ export const User = graphInvokableFactory<IUser>(_User);
 @defaultPath("users")
 @getById(User)
 @hasDelta()
-export class _Users extends _DirectoryObjects<IUserType[]> { }
```
bcameron1231, line 50: Could we use `@addable` decorator here instead of the custom `Add` method?

test/graph/chats.ts
bcameron1231, line 1: Could you add that `pnpTest()` wrapper onto the tests please :)

test/graph/messages.ts
bcameron1231, line 1: Same. Could you add that `pnpTest()` wrapper onto the tests please :)

packages/graph/messages/functions.ts
bcameron1231, line 1: looking at the docs, it looks these endpoints support some limited odata expressions
https://learn.microsoft.com/en-us/graph/api/channel-getallmessages?view=graph-rest-1.0&tabs=http#optional-query-parameters
https://learn.microsoft.com/en-us/graph/api/chats-getallmessages?view=graph-rest-1.0&tabs=http

Do you think we should potentially return a GraphCollection here, that way we can also specify some $filter and $top conditions?

packages/graph/messages/types.ts
```diff
@@ -0,0 +1,49 @@
...
+        return {
+            message: (<any>this).getById(data.id),
+            data,
+        };
+    }
+}
+export interface IMessages extends _Messages, IGetById<IMessage> { }
+export const Messages = graphInvokableFactory<IMessages>(_Messages);
+
+export interface IMessageCreateResult {
```
bcameron1231: Quick question on this: is returning data required? I thought we had previously decided to remove data/queryable and just return the promise, so just wanted to confirm whether bringing it back here was intentional.

packages/graph/teams/types.ts
```diff
@@ -334,68 +329,29 @@ export class _Channels extends _GraphCollection<IChannelType[]> {
...
-    /**
-     * Gets all the messages in a channel.
-     * @param model optionally specify the licensing and payment model
-     *
-     */
-    public async getAllMessages(model: "A" | "B" | undefined): Promise<IMessageType[]> {
-        const qString = `getAllMessages${model ? `?model=${model}` : ""}`;
-        return graphGet(GraphQueryable(this, qString));
-    }
+    // /**
```
bcameron1231: Should we remove this commented out code?

packages/graph/messages/types.ts
```diff
@@ -0,0 +1,49 @@
...
+@defaultPath("messages")
+@getById(Message)
+export class _Messages extends _GraphCollection<IMessageType[]> {
+
+    /**
+     * Adds a message
+     * @param message ChatMessage object that defines the message
+     *
+     */
+    public async add(message: IMessageType): Promise<IMessageCreateResult> {
```
bcameron1231: Can we use `@addable` here?

bcameron1231 (changes requested):
Nicely done! :) Just added a few changes for you.

juliemturner:
@bcameron1231 I think this updates everything and fixes the bug I found with pnpTest.

packages/graph/messages/functions.ts
juliemturner: Yes, absolutely I've adjusted.

packages/graph/messages/types.ts
juliemturner: Yep, for sure, adjusted.

packages/graph/messages/types.ts
juliemturner: Godd question, this is left over and I think you're right it should be removed.

packages/graph/teams/types.ts
juliemturner: Yep, WIP leftovers.

packages/graph/users/types.ts
juliemturner: Yep, adjusted

test/graph/chats.ts
juliemturner: Yep done.

test/graph/messages.ts
juliemturner: Yep, done.

bcameron1231 (approved):

PR #3355, opened 2026-06-27
Fixes lock.yml - ignore claude configurations

juliemturner:
### Category

- [x] Bug fix?
- [ ] New feature?
- [ ] New sample?
- [ ] Documentation update?

### Related Issues

lock.yml needed to be upgraded

### DIFF
--- a/.github/workflows/lock.yml
+++ b/.github/workflows/lock.yml
@@ -9,9 +9,8 @@ jobs:
   lock:
     runs-on: ubuntu-latest
     steps:
-      - uses: dessant/lock-threads@v5
+      - uses: dessant/lock-threads@v6
         with:
-          github-token: ${{ github.token }}
           issue-inactive-days: '2'
           add-issue-labels: 'auto-locked'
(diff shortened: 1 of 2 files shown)


juliemturner:
@bcameron1231 - we need to get this updated in main to.

.github/workflows/lock.yml
```diff
@@ -9,9 +9,8 @@ jobs:
   lock:
     runs-on: ubuntu-latest
     steps:
-      - uses: dessant/lock-threads@v5
+      - uses: dessant/lock-threads@v6
         with:
-          github-token: ${{ github.token }}
```
bcameron1231, line 14: Just to confirm, we don't need this anymore?

juliemturner: Correct, v6 gets around the 100 character limit  (the error in the action) by setting it by default. V6 also fixes the node20 warning as it's node24

bcameron1231 (approved):

